PR #22, opened 2018-08-10
add cf and os architecture docs

sttomm:
The documentation is currently in german as i transferred it from another place where we had this documentation in german. I think we can translate it in a second step to english.


JohannesRudolph:
we should commit the source files for the charts to the repo too so we can edit and regenerate pngs when necessary

meshjoerg:
the current state is the minimal version which needs to be expanded by some more context to make it more comprehensible to readers. accepting it for now to have a starting point and move forward.

PR #23, opened 2018-08-13
add basic description of identity federation

JohannesRudolph:


docs/meshstack.identity-federation.md
```diff
@@ -0,0 +1,35 @@
+# Identity Federation
+
+Meshstack supports integration with common enterprise single-sign-on system infrastructures. Identity Federation enables enterprises to integrate corporate SSO systems at a single point (the **Mesh Identity Broker**). Meshstack then automatically ensures identity and authorization federation into all connected **Platform Instances** as depicted below.
+
+```text
+Enterprise IdP A --> Mesh IdB -> Meshfed ---> Platform Instance X
+                  /                       \
+Enterprise IdP B -  . . . . . . . . . . .  -> Platform Instance Y
```
meshjoerg: I don't understand the meaning of the ... between IdP B and Platform Y as it seems to connect them in parallel to the IdB. I would expect a figure in the shape of a laying X. Maybe just remove the dotted line?

JohannesRudolph: hm... i intended to show that Mesh IdB and Platform Instances can participate in direct token exchange (no Meshfed needed), but I think this makes the figure more convoluted than necessary.

PR #44, opened 2018-11-13
Introduce audience personas

meshjoerg:
As discussed in today's print review, we add audience personas to focus documentation and release notes.


tfelix:
Looks good to me.

PR #49, opened 2019-01-25
Admin area

sttomm:
some sections in the table with the functionality available to the different groups is not linked to a documentation page yet. I created further tickets in Jira to extend our docs with this information.


meshjoerg:
Thanks for the update, Stefan! I made some additions or minor changes. We should try to avoid too many passive constructs and in terms of style instead of the pattern "you have to do XY" better use "to achieve X, do Y".
Thanks for the amendment!

PR #68, opened 2019-03-19
add meshApi documentation

sttomm:


JohannesRudolph:
I think it makes sense to find a way so we can extract the meshfed-api docs and include it here in the docusaurus repo. Should probably be part of the process that generates our release notes. 

Nonethelss, that's going to be a separate task

docs/meshstack.api.md
```diff
@@ -0,0 +1,6 @@
+---
+id: meshstack.api
+title: API
+---
+
+meshStack provides APIs that can be consumed by third-party applications. This allows the integration of meshStack into established company processes provided by other applications. The API documentation can be found in every meshStack installation in the footer of meshPanel. It provides all details needed to use the API. You can also find the deployed API documentation [here](https://federation.meshcloud.io/docs/index.html).
```
JohannesRudolph: f: don't hardlink to meshcloud-prod, instead include notes about relative url ?

sttomm: hmm, in that case i like the the link, because it is currently the only option we have to provide this API to new customers, who are just checking our docs for functionality we provide. So i would keep this link for now. We can change it, when we include the API docs into this public doc

PR #77, opened 2019-06-26
Adds docs for partner invite link

tfelix:


docs/meshstack.configuration.md
```diff
@@ -0,0 +1,22 @@
+---
+id: meshstack.configuration
+title: Configuration
+---
+
+Here you can find an overview of the possible configuration values in order to customize your meshStack installation.
+
+The configuration is done via a [YAML file](https://en.wikipedia.org/wiki/YAML) which is provided when the app starts.
```
JohannesRudolph: I'm not sure what's the best approach is to document these things. On one hand it's interesting for our customers/operators to see which configuration options are available but otoh we mostly deliver configured meshstack instances as a service and it's unlikely operators will poke into yaml configuration themselves. Ideas?

sttomm: i think eventhough most customers won't touch these configs, some perhaps may in the future. And knowing about the options is interesting for customers, so i think documenting it that way is ok for me.

PR #97, opened 2019-10-14
Adds AWS LZ setup documentation

tfelix:


tfelix:
This must be postponed as the details of the replication are changing.

PR #112, opened 2019-11-07
Adds AWS role config information

tfelix:


JohannesRudolph:
@tfelix can you document the configuration in terms of the dhall config model https://docs.dev.meshcloud.io/docs/meshstack.configuration.html ?

PR #114, opened 2019-11-13
Adds account alias pattern doc

tfelix:


JohannesRudolph:
@tfelix Please document this in terms of the dhall config model

tfelix:
I dont think we use/document our internal dhall config in the documentation. What do you mean exactly?

PR #115, opened 2019-11-13
fix(okd): Adds custom role description

tfelix:


docs/meshstack.openshift.index.md
```diff
@@ -217,6 +217,28 @@ oc get serviceaccount meshfed-service -o json | jq '.secrets[].name' | grep toke
 
 Operators need to securely inject this access token into the configuration of the OpenShift module.
 
+### Custom Account Roles
+
+If you want to use custom account roles (and not the pre-defined `admin`, `edit` and `view`) you need to make sure to also list these roles in the clusterrole binding section for the meshfed-service principle. It is not allowed for the service-principle to grant roles granting more rights then itself has, so the right to bind these roles must be explicitly granted.
```
JohannesRudolph: you mention "account roles" which is a term we don't use anywhere else yet - can you document this as meshProject roles and also add links to the relevant page explaining meshProject roles?

tfelix: yes also improved the wording a bit.

docs/meshstack.openshift.index.md
```diff
@@ -218,8 +218,65 @@ roleRef:
...
+  resourceNames:
+  - admin
+  - edit
+  - view
+  - my-custom-role
+```
+
 #### Metering
 
+Next, retrieve the access token for the service account:
```
sttomm: f: Everything from here on must be deleted. It is currently duplicated. LZ is now defined in separate section and the way how to access the service token has changed a bit. This was just recently merged, so something went wrong during merge or so

docs/meshstack.openshift.index.md
```diff
@@ -218,8 +218,65 @@ roleRef:
   name: meshfed-service
 ```
 
+### Custom meshProject Roles
```
sttomm: i think we should also add it to the kubernetes docs

tfelix: Good point.

PR #131, opened 2020-01-03
feat(search): add algolia search

SimonTheLeg:
Add algolia search for documentation. apiKey apparently does not have to be kept secret (example from official docusaurus config https://github.com/facebook/docusaurus/blob/1c032a0bd95b7c95e1985ff104d43416c16714e1/website/docusaurus.config.js#L74)

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -103,6 +103,11 @@ const siteConfig = {
   scrollToTop: true,
   scrollToTopOptions: {
     zIndex: 100,
+  },
+
+  algolia: {
+    apiKey: '651f9d1e5d32c2a465935a001a047407',
+    indexName: 'meshcloud_dev'
   }
 };
 


website/siteConfig.js
```diff
@@ -103,6 +103,11 @@ const siteConfig = {
   scrollToTop: true,
   scrollToTopOptions: {
     zIndex: 100,
+  },
+
+  algolia: {
+    apiKey: '651f9d1e5d32c2a465935a001a047407',
+    indexName: 'meshcloud_dev'
```
JohannesRudolph: d: how do we ensure pre-release (docs.dev.meshcloud.io) vs. release (docs.meshcloud.io) is indexed seperately? 
for now it's probably sufficient just to index prod, but both versions would be nice.

JohannesRudolph: alternatively we could use versions https://docusaurus.io/docs/en/versioning#__docusaurus and just host a single isntance of the docs, but that may be another/separate major undertaking

SimonTheLeg: oh wait I think at the moment we only have an index for docs.dev.meshcloud.io. Will request another index for PROD now

JohannesRudolph: I'm happy to merge this for now, let's add the prod instance once you have the API key for it

PR #164, opened 2020-03-26
feat: tags as cloud parameters and as chargeback statement export headers

pubudug:


JohannesRudolph:
d: do you think it makes sense to introduce a central page that talks about what kind of metadata we have available about projects and provide to Landing Zone Assets (Cloud Functions, templates, tags etc.)?

pubudug:
> d: do you think it makes sense to introduce a central page that talks about what kind of metadata we have available about projects and provide to Landing Zone Assets (Cloud Functions, templates, tags etc.)?

PR #184, opened 2020-04-29
feat: no unit costs for gcp and azure

pubudug:


docs/meshcloud.project-metering.md
```diff
@@ -47,6 +47,9 @@ From a Tenant Usage Report you can also access a Detailed Report to see every si
 
 Partner accounts can access tenant usage reports for all their assigned meshCustomers in the Administration Area via "Platforms" -> "Usage Reports".
 
+Note that Azure and GCP do not provide unit costs in their usage and billing data.
```
chewbeccy, line 50: Note that unit costs for Azure and GCP services are not supported. Therefor , the unit costs are not available in the Tenant Usage Reports for those platforms.

PR #195, opened 2020-05-20
feat: alternate payment method

pubudug:


chewbeccy:
We revisited the naming decision of the feature because the current name "alternate payment method" doesn't provide the necessary background to understand what it is. we (pubudu, stefan, rebecca) agreed to call the different payment methods applied to a project active and substitute payment method.

docs/meshcloud.project.md
```diff
@@ -34,6 +34,8 @@ When adding a [meshPlatform](meshcloud.platform-location.md) to a meshProject, m
 
 You can configure one or multiple [payment methods](meshcloud.project-metering.md#payment-methods) and [billing addresses](meshcloud.project-metering.md#company-billing-addresses) for you meshCustomer. For a meshProject you have to select a specific one, so the project resources can be billed via a specific payment method and to a specific billing address.
 
+You also have the option of setting an alternate payment method on a project. This feature is enabled via a configuration in meshStack. A job that runs daily will check if the already assigned payment method is expired and if it is, will replace it with the alternate payment method.
```
JohannesRudolph, line 37: d: you can link this into the operator documentation.

JohannesRudolph, line 37: s: Please avoid passive voice as it is unprecise "This feature is enabled via a configuration in meshStack." -> better use active voice: "Operators can [enable substitue payment method](link) via a configuration option in meshStack.

docs/meshcloud.project.md
```diff
@@ -34,6 +34,8 @@ When adding a [meshPlatform](meshcloud.platform-location.md) to a meshProject, m
...
+
```
JohannesRudolph, line 38: d: what happens after payment method rotation? Is the substitute payment method moved into theh "active payment" method slot and the "new" substitute paymenet method is then empty? Please add more details

docs/meshcloud.project.md
pubudug: I don't understand what you mean. Do you mean add a link from operator documentation to this page?

pubudug: I will resolve and merge this. Will create a different PR when I have the answers.

pubudug: I will resolve and merge this. Will create a different PR when I have the answers.

PR #208, opened 2020-06-15
feat: customer user groups viewing CU-5ynk53

pubudug:
Related to https://github.com/meshcloud/meshfed-release/pull/70. Although, this change is not about administrator rights for the Partner to view Groups within customers, but rather that the partner user can view his own groups (within the partner customer). So I did the change in meshcloud.customer.md and not administration.md.

### DIFF
--- a/docs/meshcloud.customer.md
+++ b/docs/meshcloud.customer.md
@@ -30,6 +30,11 @@ You can initially setup the role in the dropdown which describes the access leve
 Press "+" to send an invitation email. This step is necessary in order to give a user access to your [projects](meshcloud.project.md).
 After this action the created invitation will appear under the pending customer user role requests section.
 
+## View User Groups
+
+You can view user groups within your customer account by going to the **Groups** section in the **Account** area.
+Currently, creating a group is only supported via [meshObject API](meshstack.api.md).
+
 ## Assign meshCustomer Roles
 
 You can change the role assigned to each user on the current meshCustomer.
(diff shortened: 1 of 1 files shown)


meshcloud-gmbh:
Task linked: [CU-5ynk53 Update public docs (see https://github.com/meshcloud/meshfed-release/pull/70)](https://app.clickup.com/t/5ynk53)

docs/meshcloud.customer.md
```diff
@@ -30,6 +30,11 @@ You can initially setup the role in the dropdown which describes the access leve
 Press "+" to send an invitation email. This step is necessary in order to give a user access to your [projects](meshcloud.project.md).
 After this action the created invitation will appear under the pending customer user role requests section.
 
+## View User Groups
```
JohannesRudolph, line 33: should we call them "User Groups" or "meshCustomer User Groups" (which is sort of the official name)? It's clear here since we're on the meshCustomer page, but e.g. below the heading also reads "Assign meshCustomer Roles". 

@chewbeccy what do you think?

pubudug: One thing I want to raise here is that we should at least switch to user group or simply group in the description of the feature. Otherwise if we are going to write a long description with multiple paragraphs and stick with "meshCustomer User Group" I think we would be using a lot of words without adding much information.

chewbeccy: The official name should stay "meshCustomer User Group" because we maybe want to add project user groups or global user groups in the future. When the context is clear "Customer" its ok to reference them simply as user groups like @pubudug suggested.

PR #210, opened 2020-06-23
feat(gcp): Adds GCP role mapping docs

tfelix:
Also improves AWS lambda docs.

### DIFF
--- a/docs/meshstack.aws.landing-zones.md
+++ b/docs/meshstack.aws.landing-zones.md
@@ -100,6 +100,10 @@ While it is possible to trigger an AVM execution via a StackSet a better alterna
 
 > Please make sure the `MeshfedServiceRole` has the rights in order to trigger the configured Lambda.
 
+The Lambda is called synchronously, so please make sure your Lambdas execution time is kept low. If you have to do long running tasks its a good advice to call another lambda in an async fashion from the initial Lambda call, so you can return to the replication code without hitting HTTP timeouts.
+
+If the function could not be invoked because of an server error for example a permission error prevents the function of being called, this is regarded as a replication error and will fail the project replication.
+If the function itself signals an error, the error code returned by the function is currently not used but only logged for reference. This is done in order to allow the replication of the meshProject to be continued.
 
 The following parameters can be used in the Lambda function and are provided as a JSON payload like the following pattern:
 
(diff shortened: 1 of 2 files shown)


docs/meshstack.gcp.landing-zones.md
```diff
@@ -52,6 +52,12 @@ The properties of the provided configuration file will be expanded with properti
 In addition, any payment settings, project tags or customer tags are also provided to the template.
 For example, a tag named `myCustomerLabel` would be provided as a property with name `mesh-tag-my-customer-label`.
 
+## meshRole to Platform Role Mapping
+
+The meshProject roles must be mapped to GCP specific roles. Your are able to control this mapping with a Landing Zone setting. You can specifiy these mappings by adding role mappings and supplying a GCP Role. You can both use custom roles which look like `organizations/123123123123/roles/meshstack.project_developer` or predifined GCP roles like `roles/editor`.
```
chewbeccy, line 57:
```suggestion
The meshProject roles must be mapped to GCP specific roles. You are able to control this mapping with a Landing Zone setting. You can specifiy these mappings by adding role mappings and supplying a GCP Role. You can both use custom roles which look like `organizations/123123123123/roles/meshstack.project_developer` or predifined GCP roles like `roles/editor`.
```

docs/meshstack.aws.landing-zones.md
```diff
@@ -100,6 +100,10 @@ While it is possible to trigger an AVM execution via a StackSet a better alterna
 
 > Please make sure the `MeshfedServiceRole` has the rights in order to trigger the configured Lambda.
 
+The Lambda is called synchronously, so please make sure your Lambdas execution time is kept low. If you have to do long running tasks its a good advice to call another lambda in an async fashion from the initial Lambda call, so you can return to the replication code without hitting HTTP timeouts.
+
+If the function could not be invoked because of an server error for example a permission error prevents the function of being called, this is regarded as a replication error and will fail the project replication.
```
chewbeccy, line 105:
```suggestion
If the function could not be invoked because of a server error for example a permission error prevents the function of being called, this is regarded as a replication error and will fail the project replication.
```

docs/meshstack.aws.landing-zones.md
```diff
@@ -100,6 +100,10 @@ While it is possible to trigger an AVM execution via a StackSet a better alterna
...
+If the function itself signals an error, the error code returned by the function is currently not used but only logged for reference. This is done in order to allow the replication of the meshProject to be continued.
```
chewbeccy, line 106: why is the error code currently not used?

tfelix: because there was no real userstory to implement this. As soon as the function is able to get called we only display the returned response but dont stop replication on it. We could indeed e.g. stop the replication if the lambda signals us to do so but we did not went that far yet.

PR #220, opened 2020-07-23
feat(access): adapt document around access control

sttomm:
With the introduction of the principal abstraction for users and groups, documentation around access control needed to be adapted.

I also used the term "principal" in the documentation so we can apply a ubiquitous language. We decided not to use it in the UI, but i think it still makes sense in the documentation, as we have more options of describing what a principal is.

### DIFF
--- a/docs/meshcloud.project.md
+++ b/docs/meshcloud.project.md
@@ -71,7 +71,7 @@ Via "Projects" -> "Expired Access", the expired or soon to expire role assignmen
 
 ### Unassign user or group from a meshProject
 
-In the **Project Access** section you can click the `-` button in the row of a user or group to remove them from the project. The users and memebers will not be able to access this project in meshPortal and the cloud platforms anymore. You can add the  to your project again later on and all related users will get access again.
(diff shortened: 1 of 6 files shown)


docs/meshcloud.project.md
```diff
@@ -40,38 +40,38 @@ Depending on the configuration of your meshStack implementation, you may be able
...
 - **Project User**: A default user, like a developer, who can manage resources in the cloud platform.
 - **Project Admin**: An admin user, who can also change configurations of the project in the cloud platform.
 
 ### Update Project Role
 
-You can change the Project Role of a principal, by choosing a different role for the principal in the dropdown at the **Project Access** section, and clicking the **disc** icon afterwards to save the change.
+You can change the Project Role, by choosing a different role in the dropdown at the **Project Access** section, and clicking the **disc** icon afterwards to save the change.
 
-### Expiration of a principal assignment
+### Expiration of a user or group assignment
```
sttomm, line 65: d: When changing headlines, we always have to check for deep links. deep link work by using the headline in kebap-lower-case. So search for "expiration-of-a". If you find occurrences, you have to change the link accordingly. Sadly we don't get any compile errors for deep links. Errors only pop up when there is an issue in finding another md file in general.

PR #233, opened 2020-08-12
Feature/cu 5yfqfm consistently use gcp service account instead of impersonated user

JohannesRudolph:


meshcloud-gmbh:
Task linked: [CU-5yfqfm Consistently use GCP service account instead of impersonated user](https://app.clickup.com/t/5yfqfm)

PR #237, opened 2020-08-17
feat(openshift): added Service Account information for metering

younGihan:


docs/meshstack.openshift.metering.md
```diff
@@ -7,7 +7,7 @@ meshStack supports metering and billing for most common OpenShift versions.
 
 ## Prerequisites
 
-- The "meshfed-metering" service account must be configured as described [here](./meshstack.openshift.index.md#metering)
+- The "meshfed-metering" service account must be configured as described [here](./meshstack.openshift.index.md#metering-sa)
```
JohannesRudolph, line 10: d: should we setup the metering SA in the "metering" section? Or should that be part of the core platform configuration? 

I'm asking because from my PoV it makes sense to keep them separate in the docs because they're "physically" separate SPNs as well (on all our platforms)

younGihan: I understand you PoV. We could move it to the metering section - I thought if the customer needs to create both service accounts it would make sense that he could create them both at once

PR #243, opened 2020-08-19
Feature/start script

JohannesRudolph:


hack/start
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
sttomm, line 1: d: Why did you put this script in the hack folder? I'd prefer to have this in the root folder of the repo, because i would not search for a start script in the hack folder. And it doesn't look like a hack to me but just a helpful start script ;)

hack/start
```diff
@@ -0,0 +1,38 @@
...
+
+doc() {
+    cat <<-EOF
+Starts a local docusaurus development server
+
+USAGE:
+   start
+
+EXAMPLES:
+    update-snippets
```
sttomm, line 16: d: I am not sure about this example? What does that tell us? I can't see that this would be used anywhere? Same applies to `start`. Does that have any consequence? Perhaps it's just my lack of bash knowledge. But i can't see that it is actually used. just looks to me like the main function is only called if any parameter is given? Otherwise this docs is outputed?

sttomm (approved):

hack/start
JohannesRudolph: idk, it got kind of conventional in the k8s space to have a hack folder with helper scripts etc. 
I don't mind either way

hack/start
JohannesRudolph: sorry `update-snippets` is a copy paste error. yes it's a bit weird to have examples when there are no parameters but its just a conventional script scaffold to provide a help text like this

PR #246, opened 2020-08-24
feature: identity connector basic documentation

pubudug:
@sttomm added you to discuss if I've used the correct technical terms (for example should we call it identity import? we are importing groups and role bindings as well, but I couldn't come up with a better name for it)


docs/meshstack.identity-import.md
```diff
@@ -0,0 +1,6 @@
+---
+id: meshstack.ldap-identity-import
+title: LDAP Identity Import
```
sttomm, line 3: d: I am not sure whether we should mention LDAP here. Currently it is the only option we support, but in general, the Identity Import is intended to be used with multiple different "connectors"

docs/meshstack.identity-import.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+meshStack supports importing users and groups into meshStack from a source that supports the LDAP protocol, for example, Microsoft Active Directory. The `Simple Paged Results Control` is used to do a paginated LDAP query to fetch the users and groups. Because of this, it is required that the LDAP server supports the `Simple Paged Results Control`. Other than importing users and groups, meshStack also has the capability to assign a [customer role](./meshcloud.customer.md#assign-meshcustomer-roles) to the imported group. The entities read via LDAP is imported to meshStack via the [meshObject import API](./meshstack.api.md#meshobject-api).
```
sttomm, line 6: d: Can you add a link to some kind of ldap spec or so for the "Simple Paged Results Control" flag that needs to be set in ldap?

sttomm, line 6: d: I think you should also mention the regex we use to determine tags and also the option to identify partner admin role? So in general you should add the dhall snippet documentation here for the Identity Import service.

sttomm (changes requested):

docs/meshstack.identity-import.md
chewbeccy: Yes I agree. On the roadmap it is called Group Sync and ldap is only the first protocol we support with the integration service

docs/meshstack.identity-import.md
chewbeccy, line 6:
```suggestion
meshStack supports importing users and groups into meshStack from a source that supports the LDAP protocol, for example, Microsoft Active Directory. The `Simple Paged Results Control` is used to do a paginated LDAP query to fetch the users and groups. Because of this, it is required that the LDAP server supports the `Simple Paged Results Control`. Other than importing users and groups, meshStack also has the capability to assign a [customer role](./meshcloud.customer.md#assign-meshcustomer-roles) to the imported group. The entities read via LDAP are imported to meshStack via the [meshObject import API](./meshstack.api.md#meshobject-api).
```

docs/meshstack.identity-import.md
pubudug: Should I rename it to `Identity Import` or `Group Sync`?

docs/meshstack.identity-import.md
pubudug: Yes. But the dhall configuration wasn't complete the last time I looked at it. @felixzieger was going to look into it I think. But probably we will need a few changes to the dhall config to add examples. Not sure if we can consider that to be within the ticket.

felixzieger: I am adding the missing bits in [another PR](https://github.com/meshcloud/meshfed-release/pull/309). I will add dhall configuration snippets (haven't done that before, seems like a good learning opertunity). I think the easiest way would be wait until @pubudug has finished work on the docs and then start on a new branch (referencing the same ticket).

docs/meshstack.identity-import.md
chewbeccy: meshCustomer Group Synchronisation

pubudug: Changed.

docs/meshstack.identity-import.md
pubudug: Done.

chewbeccy (approved):

docs/meshstack.identity-import.md
pubudug: I will resolve this because the requirements are also volatile right now and the config can change based on the information we get from CoBa.

pubudug:
Closing because @sttomm is on vacation and the PR cannot be merged because changes were requested by him.

PR #249, opened 2020-08-25
fix: removed mesh to platform specific mappings that don't exist

pubudug:
Removed mapping of mesh term to platform specific term that doesn't exist in the system. @sttomm can you also verify that I didn't miss anything and there is no such mapping of meshCustomer to platform specific mapping for AWS, GCP and Azure?


docs/meshstack.index.md
```diff
@@ -36,6 +36,6 @@ The connections are shown in the following matrix table:
 | Cloud Foundry |              Organization               |                 Space                 |              UAA User              |                    Quota                    |
 |  Kubernetes   |                    -                    |               Namespace               |            Rolebinding             |               YAML Templates                |
 |   OpenShift   |                    -                    |                Project                |                User                |               YAML Templates                |
-|      AWS      |            Organization Unit            |                Account                |              IAM Role              |      CF StackSets / Organization Units      |
```
sttomm, line 39: d: perhaps we should somehow mention that for public cloud platforms the customer identifier is usually part of the tenant name (but it can actually be configured). Not sure whether that is already mentioned somewhere else. If not, i think it is an important information to add

docs/meshstack.index.md
```diff
@@ -36,6 +36,6 @@ The connections are shown in the following matrix table:
 | Cloud Foundry |              Organization               |                 Space                 |              UAA User              |                    Quota                    |
```
sttomm, line 36: d: Actually even in OpenStack it is optional to map the customer to a domain. At VW we have a flat domain model in place. all projects are located in the default domain. This is something that can be configured. So you can either represent a customer as a domain, or the customer-identifier will be part of the project-name

sttomm (approved):

docs/meshstack.index.md
chewbeccy: can we put (optional) in brackets next to the mapping to make this transparent?

docs/meshstack.index.md
pubudug: This is the overview page though. https://docs.meshcloud.io/docs/meshstack.index.html. I don't think that information fits here.

docs/meshstack.index.md
pubudug: Would it fit in the overview page https://docs.meshcloud.io/docs/meshstack.index.html? I think we are trying to give a high level idea about the mappings. It also says `For more details, please consult documentation of the platforms.`

sttomm: I think it makes sense to mention that it is not always reflected as a domain. It depends on the configuration. But details about this should be documented in a separate page. You'd have to figure out where it fits best

pubudug: Okay. I will do it here, but I will not try to update any OpenStack related page right now because my goal with this PR was to fix this small inconsistency.

JohannesRudolph (approved):

PR #250, opened 2020-08-27
feat(docs): add allowHierarchicalFolderAssignment config flag descrip…

tobiasweiss93:
…tion

Concerns CU-76u9kc

### DIFF
--- a/docs/meshstack.gcp.index.md
+++ b/docs/meshstack.gcp.index.md
@@ -210,8 +210,8 @@ let GcpPlatformCoreConfiguration =
       allowHierarchicalFolderAssignment:
         Configuration flag to enable or disable hierarchical folder assignment in GCP. This means
         projects can be assigned to sub folders of the defined resource manager folder in a landing zone.
-        If this config flag is disabled the replicator forces to create projects only directly under the resource manager folder.
-        Every project which is not directly attached to the resource manager folder will be then moved via gcp function to the right folder.
+        If this flag is disabled the replicator forces tenants directly under the resource manager folder.
+        Every tenant which is not directly attached to the resource manager folder will be moved there via gcp function.
     -}
       { platform : Text
       , domain : Text


meshcloud-gmbh:
Task linked: [CU-76u9kc Consider hierarchical folder structure in replication](https://app.clickup.com/t/76u9kc)

docs/meshstack.gcp.index.md
```diff
@@ -206,12 +206,17 @@ let GcpPlatformCoreConfiguration =
         The resulting string must not exceed a total length of 30 characters. Only alphanumeric + hyphen are allowed.
         We recommend that configuration include at least 3 characters of the random parameter to reduce the chance of
         naming collisions.
+
+      allowHierarchicalFolderAssignment:
```
chewbeccy, line 210: @tobiasweiss93 can you give a little bit more insight in the impact of the replication behaviour.

tobiasweiss93: I've extended the description.

docs/meshstack.gcp.index.md
```diff
@@ -206,12 +206,18 @@ let GcpPlatformCoreConfiguration =
...
+        Configuration flag to enable or disable hierarchical folder assignment in GCP. This means
+        projects can be assigned to sub folders of the defined resource manager folder in a landing zone. 
+        If this config flag is disabled the replicator force to create projects only directly under the resource manager folder.
```
chewbeccy, line 213: I thought the replicator would also move a project to the parent folder. Not only enforce this during project creation. Think about imported tenants via meshObject import.

tobiasweiss93: I've talked about this with Stefan. With this flag we only verify that a project is attached to a sub folder of the defined resource. If this is not the case then we call the update project function. After this step the project will be moved to the right place depend on the defined gcp function.

chewbeccy (approved):

PR #252, opened 2020-08-28
feat(docs): add missing meshLZ tags docs

tobiasweiss93:
Concerns CU-5gjxyv

### DIFF
--- a/docs/administration.landing-zones.md
+++ b/docs/administration.landing-zones.md
@@ -16,3 +16,7 @@ Please consult the [operator documentation](./meshstack.index.md) of the respect
 
 Disabled Landing Zones can't be assigned to projects anymore. If the Landing Zone has already been assigned to a project,
 this assignment will remain, but customer admins can no longer create new projects using this Landing Zone.
+
+## Configure Landing Zone tags
+
+Platform Operators can configure a custom Landing Zone [Tag Schema](./meshstack.tag-schema.md#tag-schemas). The schema is available during the Landing Zone creation and update. For each created Landing Zone different tags can be attached. You can tag or modify tags on a landing zone during creation and edit.


meshcloud-gmbh:
Task linked: [CU-5gjxyv meshLZ tags](https://app.clickup.com/t/5gjxyv)

docs/administration.landing-zones.md
```diff
@@ -16,3 +16,7 @@ Please consult the [operator documentation](./meshstack.index.md) of the respect
 
 Disabled Landing Zones can't be assigned to projects anymore. If the Landing Zone has already been assigned to a project,
 this assignment will remain, but customer admins can no longer create new projects using this Landing Zone.
+
+## Configure Landing Zone tags
+
+Platform Operators can configure a custom Landing Zone [Tag Schema](./meshstack.tag-schema.md#tag-schemas). The schema is available during the Landing Zone creation and update. For each created Landing Zone different tags can be attached.
```
chewbeccy, line 22: @tobiasweiss93  what do you mean by the schema is available during LZ creation and update?

tobiasweiss93: The sentence means that we show the schema during create and update. Depend on the schema we generate with ng-schema-form the form. Then we can select necessary tags.

chewbeccy: Aha lets change that into something users that are not aware of the JSON schema can also understand. Like "you can tag or modify tags on a landing zone during creation and edit.

tobiasweiss93: I've added your suggestion to the docs.

docs/administration.landing-zones.md
```diff
@@ -16,3 +16,7 @@ Please consult the [operator documentation](./meshstack.index.md) of the respect
 
 Disabled Landing Zones can't be assigned to projects anymore. If the Landing Zone has already been assigned to a project,
 this assignment will remain, but customer admins can no longer create new projects using this Landing Zone.
+
+## Configure Landing Zone tags
```
Jelledb, line 20: d: Just a question but restricted tags do not work here right? Should we mention here that the use of `restricted` is not possible?

Jelledb (approved):

PR #259, opened 2020-09-09
feature(identityconnecotr) configuration model CU-6rme8u

pubudug:


meshcloud-gmbh:
Task linked: [CU-6rme8u Sync partner admin group, platform operator group from AD to meshStack along with their tags](https://app.clickup.com/t/6rme8u)

docs/meshstack.customer-group-sync.md
```diff
@@ -4,3 +4,206 @@ title: meshCustomer Group Synchronisation
...
+
+
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Dhall Type-->
+```dhall
+let ConnectionConfiguration =
+    {-
+      url: The LDAP connection URL which would be of the form ldap://example.com:389
+      base:
+          The base from which searches should be performed, which would be of the form "dc=meshcloud,dc=io"
```
chewbeccy, line 26: Do we want to call it searches or lookup?

docs/meshstack.customer-group-sync.md
```diff
@@ -4,3 +4,206 @@ title: meshCustomer Group Synchronisation
...
+      username: The username of the user who can connect to the LDAP source to perform searches
```
chewbeccy, line 27: this is a technical user right not a personalized one. Think we should mention that somewhere

docs/meshstack.customer-group-sync.md
```diff
@@ -4,3 +4,206 @@ title: meshCustomer Group Synchronisation
...
+The configuration related to the collection of data via LDAP is as follows.
+
+<!--snippet:mesh.identityconnector.ldap.collector#type-->
+
+
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Dhall Type-->
+```dhall
+{-
+  pageSize: The size of a single page that will be returned as a result of an LDAP search
```
chewbeccy, line 43: page size is mandatory right? Can you add that please

docs/meshstack.customer-group-sync.md
```diff
@@ -4,3 +4,206 @@ title: meshCustomer Group Synchronisation
...
+  group/person:
+    a configuration of type LDAPFilter which will contain the information needed to search the LDAP source
```
chewbeccy, line 45:
```suggestion
    a configuration of type LDAP filter which will contain the information needed to search the LDAP source
```

docs/meshstack.customer-group-sync.md
```diff
@@ -4,3 +4,206 @@ title: meshCustomer Group Synchronisation
...
+
+
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Dhall Type-->
+```dhall
+let TransformationConfiguration =
+    {-
+    mesh-object:
+      The type of meshObject this configuration is about. Can be one of
+      MeshUser, MeshCustomerUserGroup or MeshCustomerGroupBinding
```
chewbeccy, line 88:
```suggestion
      meshUser, meshCustomerUserGroup or meshCustomerGroupBinding
```

chewbeccy (commented):
- f please see comments

docs/meshstack.customer-group-sync.md
tfelix: d: wouldn't it be more clearer when we maybe call it "basePath" or maybe "baseDomain"? dc stands for domain component I think, or something like that? I am not sure how the official term would be here. If you say everyone in the LDAP corner knows what a base is then its okay. Maybe we don't need to change the variable name as this is a load of work but maybe make it in the description clearer.

docs/meshstack.customer-group-sync.md
tfelix: I think Pububdu called it that way because its an actual type but I am not sure if we should bother the readers here with actual types. Maybe go the route that we only mention its an so called LDAP filter and here you have the info how to setup such a filter for our system.

docs/meshstack.customer-group-sync.md
```diff
@@ -4,3 +4,206 @@ title: meshCustomer Group Synchronisation
...
+    for users and groups.
+-}
+  let CollectorConfiguration =
+        { pageSize : Natural, group : LdapFilter, person : LdapFilter }
+
+  in  CollectorConfiguration
+```
+<!--END_DOCUSAURUS_CODE_TABS-->
+
+The LdapFilter mentioned above has the following form.
```
tfelix, line 55: t: Adapt that to the outcome of the above decision.

docs/meshstack.customer-group-sync.md
```diff
@@ -4,3 +4,206 @@ title: meshCustomer Group Synchronisation
...
+<!--snippet:mesh.identityconnector.ldap.filter#type-->
+
+
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Dhall Type-->
+```dhall
+{-
+    attributes:
+        A comma separated list of attributes that should be returned per LDAP entry.
+        For example, "cn, uniqueMember, description"
```
tfelix, line 66: d: Ah okay was this changed? I think previously we just returned the whole entity. Was there a problem with it? If I would have not code insight I would have problems to understand this text, probably mention that the returned attributes are later used for transformation into meshObject properties so the purpose of it is clearer and I can think about what properties I need?

docs/meshstack.customer-group-sync.md
```diff
@@ -4,3 +4,206 @@ title: meshCustomer Group Synchronisation
...
+<!--Dhall Type-->
+```dhall
+let TransformationConfiguration =
+    {-
+    mesh-object:
+      The type of meshObject this configuration is about. Can be one of
+      MeshUser, MeshCustomerUserGroup or MeshCustomerGroupBinding
+    fields:
+      A list of configuration objects specifying which LDAP attribute gets mapped to which
+      mesh object attribute, and whether any transformations should be done on the attribute before
```
tfelix, line 91:
```suggestion
      meshObject attribute, and whether any transformations should be done on the attribute before
```
t: think its a special name

docs/meshstack.customer-group-sync.md
tfelix: hm no the names given here I think are enum values which have this convention of being uppercase at the beginning. So that are indeed the legal values. Maybe put this into code tags so its clear we deal here not with our business names but with actual values.
```suggestion
      `MeshUser`, `MeshCustomerUserGroup` or `MeshCustomerGroupBinding`
```
oh I see its insie a dhall code block... well then good luck. :D Seems we need to keep it like this. Maybe quote it with " at least.

docs/meshstack.customer-group-sync.md
```diff
@@ -4,3 +4,206 @@ title: meshCustomer Group Synchronisation
...
+      assigning it to the meshObject field.
+    tags: Same as fields above, but for tags on the specified meshObject.
+    -}
+      { mesh-object : Text, fields : List Field, tags : List Field }
+```
+<!--END_DOCUSAURUS_CODE_TABS-->
+
+The Field configuration mentioned above has two possible forms. A basic form known as `BaseField` and a Field that performs
+transformations on the matched attribute, called the `RegexField`. The `RegexField` is an extension of the `BaseField`.
```
tfelix, line 100:
```suggestion
transformations on the matched attribute, called the `RegexField`.
```
d: think that info is not required for the reader.

tfelix (approved):

docs/meshstack.customer-group-sync.md
felixzieger: We specified attributes since the beginning. There was no explicit decision for that. 

I agree that the connection to transform part should be made clear. An alternative would be to re-write the config to only specify the attributes in the transformation configuration and infer the needed attributes for spring via dhall. This way we eliminate configuration errors where attributes are expected in transformations and not specified in collector configuration.

docs/meshstack.customer-group-sync.md
pubudug: Not sure if ActiveDirectory (or other LDAP sources) have the concept of a technical user. I could say technical user here but not sure if that would clarify or confuse.

docs/meshstack.customer-group-sync.md
pubudug: @tfelix  The spring documentation also only says `base`  so everyone in the LDAP corner would know it I think. And we also mention an example here. 
@chewbeccy  the term LDAP search is used in a lot of LDAP docs. I think lookup would mean looking up a single LDAP entry.

docs/meshstack.customer-group-sync.md
pubudug: The fact that is is mandatory can be seen from the model. If it was optional it would say `Optional Text`

docs/meshstack.customer-group-sync.md
pubudug: Yes. It is an actual type and it's structure is shown right below.

docs/meshstack.customer-group-sync.md
pubudug: It was not changed by me. When I picked up the task it was already like that.

docs/meshstack.customer-group-sync.md
```diff
@@ -4,3 +4,206 @@ title: meshCustomer Group Synchronisation
 ---
 
 meshStack supports importing users and groups into meshStack from a source that supports the LDAP protocol, for example, Microsoft Active Directory. The [`Simple Paged Results Control`](https://www.ietf.org/rfc/rfc2696.txt) is used to do a paginated LDAP query to fetch the users and groups. Because of this, it is required that the LDAP server supports the `Simple Paged Results Control`. Other than importing users and groups, meshStack also has the capability to assign a [customer role](./meshcloud.customer.md#assign-meshcustomer-roles) to the imported group. The entities read via LDAP are imported to meshStack via the [meshObject import API](./meshstack.api.md#meshobject-api).
```
felixzieger, line 6:
```suggestion
meshStack supports importing users and groups into meshStack from a source that supports the LDAP protocol, for example, Microsoft Active Directory. The [`Simple Paged Results Control`](https://www.ietf.org/rfc/rfc2696.txt) is used to do a paginated LDAP query to fetch the users and groups. Because of this, it is required that the LDAP server supports the `Simple Paged Results Control`. Other than importing users and groups, meshStack also has the capability to assign a [customer role](./meshcloud.customer.md#assign-meshcustomer-roles) or [partner role](./administration.index.md) to the imported group. The entities read via LDAP are imported to meshStack via the [meshObject import API](./meshstack.api.md#meshobject-api).
```

felixzieger: d: Mention possibility to add partner role meshCustomerGroupBinding (see suggestion)

docs/meshstack.customer-group-sync.md
pubudug: When configuring a RegexField, they would also have to include the properties of the BaseField. So I think the information is useful. They can see that in the dhall model below as well.

felixzieger (approved):

docs/meshstack.customer-group-sync.md
chewbeccy: ok when it is a standardized term I'm fine with it

docs/meshstack.customer-group-sync.md
chewbeccy: @tfelix agreed

chewbeccy (approved):

docs/meshstack.customer-group-sync.md
pubudug: Will say `technical user`. I suppose they would understand.

docs/meshstack.customer-group-sync.md
pubudug: Will add quotes.

docs/meshstack.customer-group-sync.md
pubudug: @felixzieger we could do that but shouldn't be part of this PR I think. Please report a ticket if you think it would be needed so that it can be handled seperately.

docs/meshstack.customer-group-sync.md
pubudug: Will change.

PR #267, opened 2020-09-16
docs(azure): fixed the privilege escalation policy.

younGihan:
the deleted part did not work as azure only allows 1x wildcard (*) and as it would only check "any" role the removal has no other negative impact


JohannesRudolph:
Should this also get a note that operators can only activate this policy after they've set up all the permissions for the SPNs?

JohannesRudolph (approved):

PR #275, opened 2020-10-06
feat(style): Add some basic styling for tables

SimonTheLeg:
This is important because before we did not have any styling for header
rows, making it impossible for them to be distinguished from regular
rows


JohannesRudolph (approved):
lgtm

PR #276, opened 2020-10-06
Feature/cu 8my9cr fix info on idp spp perm

SimonTheLeg:
This change fixes missing permission scope explanations and correctly links meshstack config to spp config.

There is also a separate PR for styling tables, which are used in this PR: #275. Nevertheless both PRs can be merged independently

### DIFF
--- a/docs/meshstack.configuration.md
+++ b/docs/meshstack.configuration.md
@@ -191,7 +191,7 @@ Operators can configure the mailbox this feedback is sent to via `meshfed.web`:
 
 When you add users to your [meshCustomers](./meshcloud.customer.md) we currently support live typeahead for users stored in an Azure AAD Identity Provider and Google Cloud Directory (GCD). This makes it easier for people to invite additional users without remembering their full contact details.
 
-#### Configure for Azure AAD
+#### Configure Azure AAD
 
 In order to use this functionality, create a new principal (described in **Replicator** &rarr; **AAD Level Permissions** step 1 and 2) and assign the following required permissions:
 
--- a/docs/meshstack.configuration.md
+++ b/docs/meshstack.configuration.md
@@ -193,13 +193,17 @@ When you add users to your [meshCustomers](./meshcloud.customer.md) we currently
 
 #### Configure for Azure AAD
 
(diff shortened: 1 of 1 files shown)


felixzieger (approved):

SimonTheLeg:
merge conflicts have been resolved

docs/meshstack.configuration.md
```diff
@@ -193,13 +193,17 @@ When you add users to your [meshCustomers](./meshcloud.customer.md) we currently
 
 #### Configure for Azure AAD
```
felixzieger, line 194: This should be consistent with the GCD header: with or without "for" 
```suggestion
#### Configure Azure AAD
```

felixzieger (approved):

docs/meshstack.configuration.md
```diff
@@ -191,15 +191,19 @@ Operators can configure the mailbox this feedback is sent to via `meshfed.web`:
 
 When you add users to your [meshCustomers](./meshcloud.customer.md) we currently support live typeahead for users stored in an Azure AAD Identity Provider and Google Cloud Directory (GCD). This makes it easier for people to invite additional users without remembering their full contact details.
 
-#### Configure for Azure AAD
+#### Configure Azure AAD
 
-In order to use this functionality, you must create a new principal (described in **Replicator** &rarr; **AAD Level Permissions** step 1 and 2) and assign the following required permissions as an **application permission**:
+In order to use this functionality, create a new principal (described in **Replicator** &rarr; **AAD Level Permissions** step 1 and 2) and assign the following required permissions:
```
tfelix, line 196: d: Shouldnt this whole section rather moved to https://docs.dev.meshcloud.io/docs/meshstack.configuration.html#user-identity-lookup? To avoid this double documentation. In the end this describes on how to setup IDP lookup which is only used in meshfed-api so I voted to move the docs there. Or we should remove it form there. You can decide. :D

tfelix (changes requested):

docs/meshstack.configuration.md
SimonTheLeg: yes you are right we do not need the duplication

SimonTheLeg:
closed, as this PR only would duplicate information.

PR #281, opened 2020-10-08
feat: Removes GCP domain delegation guide

tfelix:
Adds info on how to add user with group admin role.

Concerns: CU-7uw4e9

### DIFF
--- a/docs/meshstack.gcp.index.md
+++ b/docs/meshstack.gcp.index.md
@@ -100,58 +100,22 @@ In order to associate created projects with a Billing Account, the replicator ne
 
 ## Cloud Identity Configuration
 
-### meshfed-service User
+### Authorizing the Service Account
 
-The Service Account created above needs to impersonate a technical user to perform [delegated operations using the Admin SDK](https://developers.google.com/admin-sdk/directory/v1/guides/delegation).
-Depending on your organization's setup of Google Cloud Identity, provisioning a technical user in the cloud directory
-may require one of the following two alternatives.
+In order to perform certain group related administrative tasks the previosly created service account needs an additional role from the Admin Console (G Suite). In order to add this role you need to get the `unique ID` of the Service Account first.
 
-1. Provisioning the user as a "cloud only" Account in the [Google Admin Console](https://admin.google.com/).
-2. Provisioning the user in an on-premises directory synced with the cloud directory
+Please follow the official [Google guide](https://cloud.google.com/identity/docs/how-to/setup#auth-no-dwd) in order to:
 
-> In any case, you will have to log in with the user once to accept Google's Terms of Service. If you provision the user
-> as a "cloud only" account but have federation enabled on your cloud directory, temporarily making the technical user
-> a ["Super Admin" disables SSO](https://support.google.com/a/answer/6341409?hl=en) and enables cloud only login.
+1. Find the `roleID` of the `Group Administrator` role in your Admin SDK.
+2. Use the [Role Assignments API](https://developers.google.com/admin-sdk/directory/v1/reference/roleAssignments/insert) and assign this role to the service account with help of its unique id. The HTTP POST payload looks like:
 
-
-We recommend calling this user `meshfed-service`. You may also want to consider provisioning it on a domain reserved for technical
-user accounts. The user will be identified by an email like:
-
-```text
-meshfed-service@dev.meshcloud.io
+```json
+{
+  "assignedTo": "SERVICE_ACCOUNT_UNIQUE_ID",
+  "roleID": "ROLE_ID",
+  "scopeType": "CUSTOMER"
+}
 ```
-
-You can reset the password of this user after completing the setup instructions. This helps ensure that only the
- `meshfed-service` Service Account is able to impersonate the technical user account.
-
-#### Assign Admin Roles
-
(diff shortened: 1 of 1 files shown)


tfelix:
@JohannesRudolph Could you have a look if I removed the right impersonation section here? I dont think more then that is required. At least when we performed these steps access to GCP was working again.

meshcloud-gmbh:
Task linked: [CU-7uw4e9 Simplify GCP Integration without service user and Domain Wide Delegation](https://app.clickup.com/t/7uw4e9)

chewbeccy (approved):

docs/meshstack.gcp.index.md
```diff
@@ -102,56 +102,14 @@ In order to associate created projects with a Billing Account, the replicator ne
 
 ### meshfed-service User
```
JohannesRudolph, line 103: d: the title is misleading now, right? should be "Authorizing the Service Account" or something?

docs/meshstack.gcp.index.md
```diff
@@ -102,56 +102,14 @@ In order to associate created projects with a Billing Account, the replicator ne
...
 
-The Service Account created above needs to impersonate a technical user to perform [delegated operations using the Admin SDK](https://developers.google.com/admin-sdk/directory/v1/guides/delegation).
-Depending on your organization's setup of Google Cloud Identity, provisioning a technical user in the cloud directory
-may require one of the following two alternatives.
+In order to perform certain group related administrative tasks the previosly created service account needs an additional role from the Admin Console (G Suite). When you have executed the Terraform Script to setup the GCP service account its unique id is printed after the creation.
```
JohannesRudolph, line 105: d: while we want to go in the direction of using terraform platform modules also in the docs, at the moment the doc instructions are all agnostic to how operators provision these things... 

I'd remove the reference for now until we have that sorted out

docs/meshstack.gcp.index.md
```diff
@@ -102,56 +102,14 @@ In order to associate created projects with a Billing Account, the replicator ne
...
 
-1. Provisioning the user as a "cloud only" Account in the [Google Admin Console](https://admin.google.com/).
-2. Provisioning the user in an on-premises directory synced with the cloud directory
+In case you missed it, the Google guide on [how to setup a admin role service account without domain-wide delegation](https://cloud.google.com/identity/docs/how-to/setup#auth-no-dwd) also describes how to find the unique id.
```
JohannesRudolph, line 107: Leave out "In case you missed it"

docs/meshstack.gcp.index.md
```diff
@@ -102,56 +102,14 @@ In order to associate created projects with a Billing Account, the replicator ne
...
-- Go to [Manage OAuth Clients](https://support.google.com/a/answer/162106?hl=en)
-- Add a new Authorized client with these details
-  - Client Name = `Client Id` of `meshfed-service` Service Account from Google Cloud Console (displayed under “Domain-wide Delegation” on the Service Account Details Page )
-  - Scopes:
-
-      ```text
-      https://www.googleapis.com/auth/admin.directory.user, https://www.googleapis.com/auth/admin.directory.group
-      ```
+1. Find the `roleID` of the `Group Administrator` role in your Admin SDK.
+2. Use the [Role Assignments API](https://developers.google.com/admin-sdk/directory/v1/reference/roleAssignments/insert) and assign this role to the service account with help if its unique id.
```
JohannesRudolph, line 112: if its -> of its?

Probably better to provide the roleASsignment request payload as an example

```json
{
roleId: // role id from console
accountId: xxx
```

JohannesRudolph (approved):

JohannesRudolph:
Open question: how do we provide upgrade instructions? I.e. at least the release notes should mention that GCP platform operators should remvove the meshfed-service users from their domain (it's not needed anymore, right?) and set up the service account role instead. 

For a "non-breaking" upgrade the instructions should also have concrete steps
1. attach service account role as in new setup docs
2. upgrade meshStack
3. remove service user as it's no longer needed

Right?

docs/meshstack.gcp.index.md
tfelix: right makes more sense.

docs/meshstack.gcp.index.md
tfelix: rephrased it a bit and removed the sentence.

docs/meshstack.gcp.index.md
tfelix: added and fixed

docs/meshstack.gcp.index.md
tfelix: was rephrased

tfelix:
Yes but I dont think the docs should include this as there is not yet a place for upgrade instructions? The release notes howToUse field might be an option here? Can add this information there. Your assumption for the upgrade process are right. I just added this to the ticket and hoped for the meshDelivery team to communicate it to the customer. But was not so sure about this process. I add this info to the release notes of the actual PR.

JohannesRudolph:
> The release notes howToUse field might be an option here?

Yes that's the best place, let's put it there!

(and it becomes part of the documentation website that way because release notes are on the docs website)

PR #296, opened 2020-11-04
Added initial user documentation for meshPolicies

Jelledb:


docs/meshcloud.policies.md
```diff
@@ -0,0 +1,64 @@
...
+  
+### When am I impacted by meshPolicies?
+
+> If your organization has no policies defined at all, the information below is not relevant.
+
+Note that it depends if your organization uses policies and on which meshObjects they are applied. We will discuss all possible places here.
+You may encounter policies in the meshPanel when doing any of the following actions:
+
+1. Creating a new project
+    1. When adding a meshLandingZone to a new project, all policies are evaluated between a 'meshProject' and a 'meshLandingZone'
```
sttomm, line 18:
```suggestion
    1. When adding a meshPlatform with a meshLandingZone to a new project, all policies are evaluated between a 'meshProject' and a 'meshLandingZone'
```

docs/meshcloud.policies.md
```diff
@@ -0,0 +1,64 @@
...
+    2. Upon saving a new project, all policies between 'meshCustomer' and 'meshProject' are evaluated. (read use cases below to know why this is useful
+2. Editing a project
+    1. When adding a new location with a meshLandingZone all policies are evaluated between the 'meshProject' and the selected 'meshLandingZone'.
+    2. When adding a new meshCustomerUserGroup to a meshProject, all policies are evaluated between the 'meshProject' and the 'meshCustomerUserGroup'.
+    3. When changing a tag value (e.g. changing the environment) of a project, **all** policies related to the meshProject are evaluated as it impacts many meshObjects. The following meshObjects will be evaluated on to the project:
+        - the meshCustomer the project lives in
+        - all assigned meshLandingZones
+        - all assigned meshUsers
```
sttomm, line 26: d: Is this actually already the case at the moment?

docs/meshcloud.policies.md
```diff
@@ -0,0 +1,64 @@
...
+        - all attached meshCustomerUserGroups
+3. Adding a meshCustomerUserGroup to a meshCustomer. All policies between 'meshCustomer' and 'meshCustomerUserGroup' are evaluated on the given meshCustomer and 'meshCustomerUserGroup'.
```
sttomm, line 28:
```suggestion
3. Adding a meshCustomerUserGroup to a meshCustomer. All policies between 'meshCustomer' and 'meshCustomerUserGroup' are evaluated on the given 'meshCustomer' and 'meshCustomerUserGroup'.
```

docs/meshcloud.policies.md
```diff
@@ -0,0 +1,64 @@
...
+
+It might happen, consciously or unconsciously, that you violate one or more policies. At every place in the meshPanel where this is possible, we prevent you from finalizing the violation and you will be prompted with an error message explaining which meshPolicy or policies you violated, and why.
+
+Take this policy violation as an example (see the picture below), where we have a meshPolicy defined on 'meshCustomer' and 'meshProject', both on the `environment` tag.
+
+![Example Policy Error Message](assets/mesh_policies/policies_example_error_message.png)
+
+The project that is being created called 'my-example-project-prod' wants environment `prod`  and the meshCustomer 'managed-customer' the project is being created in, has environments `dev` , `test` and `qa` defined. This means there is a mismatch as `<prod>` is not inside `<dev, test, qa>`. In order to solve this problem, we have to pick an environment that is defined on the meshCustomer, e.g. `dev` or `qa`. After picking either of these we can save the project again, and (if we don't violate any further policies) the project is successfully created, including the right compliance for your organization! ✅
```
sttomm, line 38:
```suggestion
The project that is being created called 'my-example-project-prod' wants environment `prod` and the meshCustomer 'managed-customer' the project is being created in, has environments `dev`, `test` and `qa` defined. This means there is a mismatch as `<prod>` is not inside `<dev, test, qa>`. In order to solve this problem, we have to pick an environment that is defined on the meshCustomer, e.g. `dev` or `qa`. After picking either of these we can save the project again, and (if we don't violate any further policies) the project is successfully created, including the right compliance for your organization! ✅
```

as it is hard to see, i just removed to unnecessay blanks with this suggestion

sttomm (approved):

docs/meshcloud.policies.md
Jelledb: Yup just checked on panel dev and it works.

docs/meshcloud.policies.md
```diff
@@ -0,0 +1,64 @@
...
+    3. When changing a tag value (e.g. changing the environment) of a project, **all** policies related to the meshProject are evaluated as it impacts many meshObjects. The following meshObjects will be evaluated on to the project:
+        - the meshCustomer the project lives in
+        - all assigned meshLandingZones
+        - all assigned meshUsers
+        - all attached meshCustomerUserGroups
+3. Adding a meshCustomerUserGroup to a meshCustomer. All policies between 'meshCustomer' and 'meshCustomerUserGroup' are evaluated on the given meshCustomer and 'meshCustomerUserGroup'.
+
+### What happens when I violate a meshPolicy?
+
+It might happen, consciously or unconsciously, that you violate one or more policies. At every place in the meshPanel where this is possible, we prevent you from finalizing the violation and you will be prompted with an error message explaining which meshPolicy or policies you violated, and why.
```
chewbeccy, line 32: As a partner I can change tags on a LZ which results in violated policies. So we dont restrict every user from moving forward with changes that will lead to non-compliance

docs/meshcloud.policies.md
```diff
@@ -0,0 +1,64 @@
...
+
+![Example Policy Error Message](assets/mesh_policies/policies_example_error_message.png)
+
+The project that is being created called 'my-example-project-prod' wants environment `prod`  and the meshCustomer 'managed-customer' the project is being created in, has environments `dev` , `test` and `qa` defined. This means there is a mismatch as `<prod>` is not inside `<dev, test, qa>`. In order to solve this problem, we have to pick an environment that is defined on the meshCustomer, e.g. `dev` or `qa`. After picking either of these we can save the project again, and (if we don't violate any further policies) the project is successfully created, including the right compliance for your organization! ✅
+
+### Are there any other places where policies are enforced?
+
+Besides end-users being impacted at the places above, there are also other places where policy violations could be created. The diagram below describes all possible relationships between meshObjects and the behavior that is expected, depending on the direction of the change.
```
chewbeccy, line 42:
```suggestion
Besides end-users being impacted at the places above, there are also other places where policy violations could be caused. The diagram below describes all possible relationships between meshObjects and the behavior that is expected, depending on the direction of the change.
```

docs/meshcloud.policies.md
```diff
@@ -0,0 +1,64 @@
+---
+id: meshcloud.policies
+title: meshPolicies
+---
+
+## Introduction: what are meshPolicies?
+
+A meshPolicy is a set of rule(s) between two [meshObjects](meshcloud.index.md#introduction) that are globally defined by your organization. Each rule describes what [tags](meshcloud.tag-schema.md) have to match on each meshObject, in order to comply with the meshPolicy. meshPolicies are enforced in various places when attaching meshObjects.
```
chewbeccy, line 8: Can we add an explanation what "match" means in this context. Something like the tags of one objects are a complete match or a subset of the tags of the comparing objects.

Jelledb: Removed the word match and replaced it with a more concrete explanation. I don't think this is the place to expand on subset tag matching, also because we don't even support this yet.

chewbeccy (approved):

PR #297, opened 2020-11-05
feat: Adds kraken GCP rights

tfelix:
Adds additional rights for GCP tenant billing.

### DIFF
--- a/docs/meshstack.gcp.index.md
+++ b/docs/meshstack.gcp.index.md
@@ -28,7 +28,7 @@ We recommend using [externally provisioned identities](./meshstack.identity-fede
 
 Operators need to setup a GCP Organization to be used by meshStack. Please review the official GCP documentation on [creating and managing organizations](https://cloud.google.com/resource-manager/docs/creating-managing-organization).
 
-### meshfed-service IAM Role
+### Service Account Configuration
 
 meshStack needs a well-defined set of permissions for its automation. meshStack is designed so that it **does not require
 access to workload**. We highly recommend that permissions are configured according to the principle of least privilege.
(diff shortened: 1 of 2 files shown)


docs/meshstack.gcp.index.md
```diff
@@ -28,7 +28,7 @@ We recommend using [externally provisioned identities](./meshstack.identity-fede
 
 Operators need to setup a GCP Organization to be used by meshStack. Please review the official GCP documentation on [creating and managing organizations](https://cloud.google.com/resource-manager/docs/creating-managing-organization).
 
-### meshfed-service IAM Role
+### Service Account Configuration
 
 meshStack needs a well-defined set of permissions for its automation. meshStack is designed so that it **does not require
 access to workload**. We highly recommend that permissions are configured according to the principle of least privilege.
```
chewbeccy, line 34:
```suggestion
access to workload**. We highly recommend that permissions are configured according to the "least privilege" principle.
```

docs/meshstack.gcp.metering.md
```diff
@@ -16,6 +16,15 @@ roles/bigquery.jobUser
 roles/bigquery.dataViewer
 ```
 
+If you want to use the billing module of meshStack you also need another service principal, which is able to discover and list projects. We recommend to also create its own IAM role and assign it to the service account together with the following permissions:
```
chewbeccy, line 19:
```suggestion
To enable the billing module of meshStack a service principal, which is able to discover and list projects, needs to be defined. We recommend to also create its own IAM role and assign it to the service account together with the following permissions:
```

chewbeccy (approved):
Made some suggestions. We should not address users directly with "you". But to be honest I dont know if this is currently already the case in some parts of the documentation.

tfelix:
probably not but is fine with me.

PR #304, opened 2020-11-26
feat: Adds the lz cloud tenant tags desc

tfelix:


docs/meshstack.tag-schema.md
```diff
@@ -267,6 +267,7 @@ The following tags values can be used in such a tag definition configuration:
 | -------------------------- | :------------------------------------------------------------------------------------------------------------------------------------------------ |
 | `${projectIdentfier}`      | The meshProject identifier                                                                                                                        |
 | `${customerIdentifier}`    | The meshProject's customer identifier                                                                                                             |
+| `${landingzone}`           | The name of the applied [Landing Zone](meshcloud.landing-zones.md).                                                                               |
```
sttomm, line 270: this is ok, if you remove the "no-landingzone" default handling in the implementation. Otherwise you have to mention here that that is set in case there is no landing zone. But i'd really prefer to not set the tag in that case.

sttomm (approved):

docs/meshstack.tag-schema.md
tfelix: To avoid question I actually left that out since people would start to ask "what you can create projects without a LZ?" which is actually not really possible. Only by accident. But I can add it.

sttomm: hmm, ok, your argumentation makes sense i think. But you have to consider that for some platforms we don't have landing zones .... ok, we also don't replicate tags for them. And as soon as we support tags we may also have landing zones?

PR #306, opened 2020-12-01
fix(azure): Adds info about BP names in LZ

tfelix:


docs/meshstack.azure.landing-zones.md
```diff
@@ -25,7 +25,7 @@ Operators can optionally define one or more [Blueprints](https://docs.microsoft.
 
 ### Blueprint Name
 
-The name of the Blueprint which should get assigned to the project. If left empty, meshProject replication will not create any Blueprint assignments.
+The name of the Blueprint which should get assigned to the project. If left empty, meshProject replication will not create any Blueprint assignments. Changing the name of the Blueprint will also change the resulting Blueprint Assignment name in Azure. If you change the name of a Blueprint e.g. to point to another Blueprint, this will result in two assignments. The old assignment is not automatically removed from the Subscription and must be removed manually.
```
chewbeccy, line 28:
```suggestion
Is the name of the Blueprint which gets assigned to the tenant. If left empty, meshProject replication will not create any Blueprint assignments. Changing the name of the Blueprint will also change the resulting Blueprint Assignment name in Azure. If you change the name of a Blueprint e.g. to point to another Blueprint, this will result in two assignments. The old assignment is not automatically removed from the Subscription and must be removed manually.
```

chewbeccy (approved):
-s

docs/meshstack.azure.landing-zones.md
sttomm: I think the "Is" at the beginning in your suggestion reads a bit strange. I think the sentence should actually start with "The". Or "Blueprint Name is the ...."

sttomm (approved):

docs/meshstack.azure.landing-zones.md
tfelix: yes agree with Stefan.

tfelix: I start it with `The`

PR #309, opened 2020-12-14
added meshUser/Group policies & user default tags

sttomm:
- also did some clean ups in file names for "tag" pages

### DIFF
--- a/docs/administration.mesh-policies.md
+++ b/docs/administration.mesh-policies.md
@@ -1,13 +1,13 @@
 ---
-id: meshstack.mesh-policies
-title: Mesh Policies
+id: administration.mesh-policies
+title: meshPolicies
 ---
 
-meshPolicies can be set up between meshObjects, i.e., meshCustomer, meshProject, meshLandingZone and
-meshCustomerUserGroup. In the administration area, there is an option of 'Policies' in the 'Compliance' section of the
-navigation area. After clicking policies, you can view all policies in your meshStack instance.
+meshPolicies are described in detail [here](meshcloud.policies.md). They can be set up between meshObjects, i.e., meshCustomer, meshProject,
+meshLandingZone and meshUser/Group. In the administration area, there is an option of 'Policies' in the 'Compliance'
+section of the navigation area. After clicking policies, you can view all policies in your meshStack instance.
 
-## Create Policies
+## Create meshPolicies
 
 Policy creation is only possible for a user who has a partner admin role. There is a 'Create Policy' button top
 right on the policies screen. After clicking on it, there is a redirection to the screen of "Create Policy" where all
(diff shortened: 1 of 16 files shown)


docs/administration.mesh-policies.md
```diff
@@ -1,13 +1,13 @@
...
-id: meshstack.mesh-policies
-title: Mesh Policies
+id: administration.mesh-policies
+title: meshPolicies
 ---
 
-meshPolicies can be set up between meshObjects, i.e., meshCustomer, meshProject, meshLandingZone and
-meshCustomerUserGroup. In the administration area, there is an option of 'Policies' in the 'Compliance' section of the
-navigation area. After clicking policies, you can view all policies in your meshStack instance.
+meshPolicies are descibed in detail [here](meshcloud.policies.md). They can be set up between meshObjects, i.e., meshCustomer, meshProject,
```
Jelledb, line 6: typo: descibed

docs/meshcloud.policies.md
```diff
@@ -5,8 +5,18 @@ title: meshPolicies
...
+
+meshUsers exist on a global level and are not related to a specific meshCustomer. If you want to define a meshPolicy to only provide certain meshUsers/Groups access to a e.g. production projects, the following aspects have to be considered:
+
+1. If a user is globally defined to have access to any production project, you can add the production environment tag to that user, so this user gets access to all production projects in all meshCustomers.
```
Jelledb, line 16: This is not possible right? All users have the same tags?

docs/meshcloud.policies.md
```diff
@@ -5,8 +5,18 @@ title: meshPolicies
...
+2. If you want to maintain per meshCustomer who has access to production projects, you have to use [meshCustomerUserGroups](./meshcloud.customer.md#user-groups) for that. IT is not possible to assign a single meshUser certain tags within a certain meshCustomer.
```
Jelledb, line 17: IT => It

docs/meshstack.meshmarketplace.metering.md
```diff
@@ -130,31 +130,6 @@ The line items for the seller will appear in the usage reports and chargeback st
 
 This feature can be used for services for which usage should be tracked but the chargeback process is not yet completely established.
 
-<!--snippet:mesh.kraken.meshMarketplace-->
```
Jelledb, line 133: Was this diff intended to be in this PR?

Jelledb (changes requested):
I'll wait for your responses and then approve afterwards :-)

docs/meshcloud.policies.md
sttomm: actually this is now possible as i added the option to define tags on meshUsers via meshObject import. The important thing is, that it is globally defined, so it affects all customers. That's what i described here. This doesn't match the CoBa usecase, but it is still a valid usecase in other scenarios imo.

docs/meshstack.meshmarketplace.metering.md
sttomm: no, i just wasn't on the latest state of develop and therefore snippet creation failed and i just removed it here. I reverted this change

docs/meshcloud.policies.md
Jelledb: I'm not questioning if it's good or not but just if it exists, I wasn't aware this is now possible so forget my comment ;-)

sttomm: actually i was questioning myself whether it is good to actually provide setting tags on users via meshObject API, but in the end i liked it and i needed for some scenario tests :P

Jelledb (approved):

PR #310, opened 2020-12-17
refactoring before introducing LZ fees

pubudug:
This is just the refactoring. I will do the LZ fees docs on a follow up so that it's easier to review.


sttomm (approved):

pubudug:
@chewbeccy I merged it because this is just the refactoring we discussed on slack. If you have any comments I can fix them in the follow up PR

PR #317, opened 2021-01-08
doc: update snippets with first-period entry

malhussan:


docs/meshstack.billing-configuration.md
```diff
@@ -362,8 +362,14 @@ let Statements =
         It is the offset of days after which chargeback statements are generated.
         This time should always be set to a higher value than the finalizeReportsAfterDays
         that are configured for kraken-worker, as only finalized reports are considered in chargeback statements.
+
+      first-period:
+        The chargeback statment period generation start date (UTC datetime format: yyyy-MM-ddTHH:mm:ssZ)
```
sttomm, line 367:
```suggestion
       Chargeback statement periods will be created monthly starting from this date on (UTC datetime format: yyyy-MM-ddTHH:mm:ssZ)
```

docs/meshstack.billing-configuration.md
```diff
@@ -377,13 +383,15 @@ let example
         , "paymentAmount"
         ]
       , period-offset-days = 10
+      , first-period = "2021-01-01T00:00:00Z"
```
sttomm, line 386: d: You removed the config for VW and EnBW now completely? But at VW it must be set to "2020-12-01" and at enbw it must remain at "2020-01-01", because otherwise we will generate more periods of the past, which we don't want. Why did you remove the configs completely? Was there some confusion coming up due to our recent discussions around it?

sttomm (changes requested):

docs/meshstack.billing-configuration.md
malhussan: Sorry for the confusion; this is the example part in the docs only. In the original PR https://github.com/meshcloud/meshfed-release/pull/840 VW is changed to 2020-12-01 and EnBW will remain using the default `application.yml` which is still on 2020-01-01.

docs/meshstack.billing-configuration.md
malhussan: Done

docs/meshstack.billing-configuration.md
malhussan: I updated the example to also be similar to `application.yml` values

sttomm (approved):

PR #332, opened 2021-02-18
Add link to api docs.

henryde:
grafik


Jelledb (approved):
Approving already, but I guess we need a CI run on develop first before this link is actually reachable, right?

henryde:
> Approving already, but I guess we need a CI run on develop first before this link is actually reachable, right?

I tested it so it's already available on dev: https://docs.dev.meshcloud.io/api/

PR #345, opened 2021-03-18
feat(docs): added federated idp comment

tobiasweiss93:
Concerns CU-ex4ae5

### DIFF
--- a/docs/meshstack.openstack.index.md
+++ b/docs/meshstack.openstack.index.md
@@ -48,3 +48,8 @@ Project Users get the following OpenStack roles:
 The actual access rights associated with these roles are managed by the OpenStack instance and are not part of meshStack.
 
 More details about the User Federation with OpenStack can be found [here](meshstack.openstack.index.md).
+
+## Federated Idp
+
+meshIdB is not required for the integration anymore. You can directly integrate your company-wide IdP to OpenStack. meshStack will make sure that users have access to the OpenStack projects they are assigned to in meshStack.
+Users which are not existing yet will be created in OpenStack and assigned to the specific groups. We create for each project an own group.
--- a/docs/meshstack.openstack.index.md
+++ b/docs/meshstack.openstack.index.md
@@ -53,4 +53,31 @@ More details about the User Federation with OpenStack can be found [here](meshst
 
 meshIdB is not required for the integration anymore. You can directly integrate your company-wide IdP to OpenStack. To do this its necessary to enable the federatedIdp within the OpenStack plaform configuration in meshfed. Besides that the Idp needs to be fully integrated within OpenStack. This means users should have access to OpenStack independent of the panel login. During the automated replication meshStack will make sure that users have access to the OpenStack projects they are assigned to in meshStack. Every new created User in OpenStack will be assigned to the MeshUsers domain to prevent the creation of one user multiple times. If the user still exists in OpenStack then the automatisation will pick this user to assign him to the specific groups. We create for each project an own group.
 
+To enabled the federatedIdp its necessary to set the property 'federated-idp-enabled' to True within the dhall configuration.
+For example:
+
+```dhall
+...
+let platformConfig =
+      Platform.OpenStack::{
+      , platform = "osussuri.eu-de-central"
+      , endpoint =
+          "https://os-ussuri.germanywestcentral.cloudapp.azure.com:5000/v3"
+      , username = "admin"
+      , password = Secret.fromEnv "OS_USSURI_PASSWORD"
+      , domain = "Default"
+      , project = "admin"
+      , use-domains = True
+      , system-user-domain = "default"
+      , default-added-roles = [ "_member_", "heat_stack_owner", "creator" ]
+      , default-added-system-users = [] : List Text
+      , idp-provider = "aad"
+      , idp-protocol = "openid"
+      , federated-idp-enabled = True
+      }
+...
+```
+
 If the federatedIdp was enabled then only the cli access screen will be available within the panel. All other OpenStack related screens are disabled.
+
+**Note:** This mode is an alternative to the previously described meshIDB integration. And it will be the superior integration in future. The old one will be removed in future.
--- a/docs/meshstack.openstack.index.md
+++ b/docs/meshstack.openstack.index.md
@@ -35,7 +35,9 @@ If the KToken expires, the OIDC/Keystone Token Exchange (steps 6-8) must be exec
 
 ![OpenStack Communication](assets/os-communication.png)
 
-## Authorization
+## Authentication & Authorization
+
+## meshIdP
 
 As OpenStack has to scope authorization to a specific project, the token from the meshIdB is also scoped to a specific project in the request. The project information can then be read from the **MC_PROJECTS** claim in the token provided by the meshIdB. OpenStack maps this attribute to a local Group associated with the project, that contains the previously mentioned roles.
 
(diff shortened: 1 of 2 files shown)


meshcloud-gmbh:
Task linked: [CU-ex4ae5 Replicate AuthZ user information to OpenStack](https://app.clickup.com/t/ex4ae5)

docs/platforms.openstack.md
```diff
@@ -23,3 +23,7 @@ meshcloud allows creation of [Service Users](meshcloud.service-user.md) for Open
 ## Quota Management
 
 Partners can enforce detailed per-project quotas for OpenStack via meshcloud.
+
+## Federated Idp
+
+meshIdB is not required for the integration anymore. You can directly integrate your company-wide IdP to OpenStack. meshStack will make sure that users have access to the OpenStack projects they are assigned to in meshStack.
```
sttomm, line 29: d: Perhaps this has to be moved to the OpenStack Platform Integration page in the operators section of the docs? Or is it that already? I think we have to explain a bit more in here. So stuff like that we create users if they don't exist yet, that we set the federated attribute accordingly. Link the OpenStack docs that explain this. Describe that we currently create one group per project and assign the users to that group.

sttomm (changes requested):

docs/platforms.openstack.md
tobiasweiss93: Yes. This makes sense. We should move this part to `meshtack.openstack.index.md`.
 I've extended the description a little bit.

docs/meshstack.openstack.index.md
```diff
@@ -48,3 +48,8 @@ Project Users get the following OpenStack roles:
 The actual access rights associated with these roles are managed by the OpenStack instance and are not part of meshStack.
 
 More details about the User Federation with OpenStack can be found [here](meshstack.openstack.index.md).
+
+## Federated Idp
+
+meshIdB is not required for the integration anymore. You can directly integrate your company-wide IdP to OpenStack. meshStack will make sure that users have access to the OpenStack projects they are assigned to in meshStack.
```
sttomm, line 54: d: I think still relevant information is missing. What are the things our customers have to consider to make this work? Explain the configuration needed in meshStack. And what needs to be done in OpenStack to actually match the users? This is at least configure the mapping to write users to the MeshUsers domain. Also explain here that in general all these users are put into that domain.

And provide a hint that the OpenStack screens in meshStack are not available in that mode.

Please think about everything that is needed for delivery or our customer to set up this feature successfully.

You should also explain that we create users if they don't exist and pick up existing ones if they already logged in. But only under the condition of them being present in the MeshUsers domain.

sttomm (changes requested):

docs/meshstack.openstack.index.md
tobiasweiss93: The part that users will be created if they not exists describes the last line.
I'll integrate your other points.

docs/meshstack.openstack.index.md
```diff
@@ -48,3 +48,9 @@ Project Users get the following OpenStack roles:
 The actual access rights associated with these roles are managed by the OpenStack instance and are not part of meshStack.
 
 More details about the User Federation with OpenStack can be found [here](meshstack.openstack.index.md).
+
+## Federated Idp
+
+meshIdB is not required for the integration anymore. You can directly integrate your company-wide IdP to OpenStack. To do this its necessary to enable the federatedIdp within the OpenStack plaform configuration in meshfed. Besides that the Idp needs to be fully integrated within OpenStack. This means users should have access to OpenStack independent of the panel login. During the automated replication meshStack will make sure that users have access to the OpenStack projects they are assigned to in meshStack. Every new created User in OpenStack will be assigned to the MeshUsers domain to prevent the creation of one user multiple times. If the user still exists in OpenStack then the automatisation will pick this user to assign him to the specific groups. We create for each project an own group.
```
sttomm, line 54: d: But how does meshStack have to be configured? If meshDelivery looks into the docs they will not know and ask us. Therefore it is important to actually show what needs to be configured. So we need an according dhall snippet in here that shows that you have to set this flag to activate it and perhaps more if there is more to configure. 

And on OpenStack side it would be helpful to have a snippet of the mapping that must be done there. You can request that from @henryde.

And in general explain that this mode is an alterantive to the previously described meshIDB integration. And it will be the superior integration in future. The old one will be removed in future. Therefore below ## Authorization there should be two sections. meshIdP and Enterprise IdP or similar. Because these are the two options that can be used for AuthZ

sttomm (changes requested):

docs/meshstack.openstack.index.md
tobiasweiss93: I don't agree that we need now a separate section. I think its ok to mention that this is an alternative against the old one. 
I've append the dhall configuration.

docs/meshstack.openstack.index.md
```diff
@@ -48,3 +48,36 @@ Project Users get the following OpenStack roles:
...
```
sttomm, line 54:
```suggestion
You can directly integrate your company-wide IdP to OpenStack. This simplifies the integration with meshStack compared to the meshIdB integration that is explained above. Using the company-wide IdP must be enabled in meshStack for each affected OpenStack platform.
```
at this place you should show how to configure it. The cleanest solution for it would be actual documentation snippets in dhall. But as we don't have that for the OpenStack configuration at all till now, it would be a bit out of scope for now as you would have to describe every config option in the OpenStack platform config. So i'd recommend to use the snippet you added below, but reduce it to only the relevant config attributes. Just mention those that are relevant for the federated IdP configuration. That are federated-idp-enabled, idp-provider, idp-protocol.

```suggestion
In OpenStack you have to configure the company-wide IdP as described here (link to OpenStack docs, ask @henryde). Make sure to configure the MeshUsers domain as the domain of all users coming from the IdP like this (get the example from @henryde).

During replication, meshStack will make sure that users have access to the OpenStack projects they are assigned to in meshStack. Users will be assigned to the according groups (per meshProject) meshStack creates in OpenStack. If 
meshStack is about to assign a user that does not exist in OpenStack yet, meshStack will create this user with the according federatedIdP attributes and create the user in `MeshUsers` domain. That way once the user logs in via the IdP, he will be mapped to the user created by meshStack. If the user already exists in OpenStack during replication, meshStack will pick up this user to assign him to the according groups.
```

docs/meshstack.openstack.index.md
```diff
@@ -48,3 +48,36 @@ Project Users get the following OpenStack roles:
...
+      , default-added-roles = [ "_member_", "heat_stack_owner", "creator" ]
+      , default-added-system-users = [] : List Text
+      , idp-provider = "aad"
+      , idp-protocol = "openid"
+      , federated-idp-enabled = True
+      }
+...
+```
+
+If the federatedIdp was enabled then only the cli access screen will be available within the panel. All other OpenStack related screens are disabled.
```
sttomm, line 81:
```suggestion
**Note** If federatedIdp is enabled, only the cli access screen will be available within the panel. All other OpenStack related screens are disabled, as personalized access from meshPanel to OpenStack is not possible with this integration approach at the moment.
```

docs/meshstack.openstack.index.md
```diff
@@ -48,3 +48,36 @@ Project Users get the following OpenStack roles:
...
+
+**Note:** This mode is an alternative to the previously described meshIDB integration. And it will be the superior integration in future. The old one will be removed in future.
```
sttomm, line 83:
```suggestion
**Note:** This mode is an alternative to the previously described meshIdB integration. And it will be the superior integration in future. The old one will be removed in future.
```

sttomm (changes requested):

docs/meshstack.openstack.index.md
tobiasweiss93: Fixed

docs/meshstack.openstack.index.md
tobiasweiss93: Fixed

docs/meshstack.openstack.index.md
tobiasweiss93: All information are necessary so I'll leave the example like it is.
MeshUsers are not "Henry" related. We'll create the MeshUsers domain during the ReplicateFederatedUsersStep. So I don't aggree to add an example.
I've asked henry about the os config. I've added the docs link.

sttomm: I actually do think that they are "Henry" related, as Henry had to configure something in OpenStack, so users are actually put into the MeshUsers domain when they log in to OpenStack. Otherwise they were alway created in the default domain and not mapped to the users we created

docs/meshstack.openstack.index.md
sttomm: I think it still makes sense to have a "Authentication & Authorization" section with 2 sub-sections "meshIdP" and "Enterprise IdP" or similar.

docs/meshstack.openstack.index.md
sttomm: And regarding the dhall config ... the problem i see with mentioning all attributes is, that it is mentione din teh federatedIdP section of the docs and e.g. "default-added-roles" or the credentials are not related to that topic at all. So if you mention them here, you also have to explain what they are actually used for. But that doesn't belong to the federated IdP section and therefore doesn't belong here.

The cleanest way would be to have a complete integration section that describe the complete platform configuration with all attributes. That one would use our dhall snippet approach to cleanly document our dhall config. And then you would only link to that section from the federated IdP section and mention that the 3 attributes mentioned before are relevant for the federated IdP config. But that change is out of scope for this PR imo and therefore i'd go with just mentioning the 3 attributes that are actually related to federatedIdP config.

docs/meshstack.openstack.index.md
```diff
@@ -48,3 +48,41 @@ Project Users get the following OpenStack roles:
...
+      , default-added-roles = [ "_member_", "heat_stack_owner", "creator" ]
+      , default-added-system-users = [] : List Text
+      , idp-provider = "aad"
+      , idp-protocol = "openid"
+      , federated-idp-enabled = True
+      }
+...
+```
+
+To enabled the federatedIdp its necessary to set the property 'federated-idp-enabled' to True within the dhall configuration.
```
sttomm, line 80:
```suggestion
To enable the federatedIdp it's necessary to set the property 'federated-idp-enabled' to True within the dhall configuration.
```

sttomm (approved):
i gave some more input on the last 2 open topics. You can decide, based on that input whether and how you will integrate it. Besides that the PR is now lgtm

docs/meshstack.openstack.index.md
tobiasweiss93: Ok. I'll remove the other lines.

docs/meshstack.openstack.index.md
tobiasweiss93: Fixed

docs/meshstack.openstack.index.md
tobiasweiss93: I've added this.

PR #348, opened 2021-04-13
fix: Document seller info CU-aew1dw

pubudug:


meshcloud-gmbh:
Task linked: [CU-aew1dw Document required `sellerInfo` configuration (metering)](https://app.clickup.com/t/aew1dw)

docs/meshstack.azure.metering.md
```diff
@@ -5,3 +5,28 @@ title: Metering
 
 meshStack imports metering data from Azure via the [Azure Cost Management API](https://docs.microsoft.com/en-us/rest/api/cost-management/). It collects data for the previous and the current month.
 Azure only provides data for the previous day, so Azure Usage Reports shown in meshStack will not provide data for the current day.
+
+## Configuring Seller Information
```
sttomm, line 9: d: We now have 3 times basically the same section. I'd prefer to have it described in general on e.g. https://docs.meshcloud.io/docs/meshstack.billing-configuration.html . And then just link to that section from the platform specific pages, if that is necessary at all.

sttomm (changes requested):

docs/meshstack.azure.metering.md
pubudug: This only applies for Azure,Aws and Gcp. And the example is different for each platform, because whoever creates the product in the db has to put the exact platformType in the correct case.

pubudug: @sttomm given this information are you still for moving the block? :)

sttomm: hmm, ok, i thought it should be available for all platforms. Then i'd just mention the specifics like all the available platformTypes. But as it is only available for public cloud platforms we may also go with the separate sections. But i'd still be in favor of putting it centrally, listing all platformTypes and mentioning that it is only available for public cloud platforms atm (but in future i guess we will also support it for private cloud platforms .... but don't we already do that? I thought we also have it configured for e.g. OpenStack at VW.

pubudug: For other platforms, we don't need a `sellerInfo` prodcut, because those platforms have actual products in our db that are used in the calculations. With the public clouds we need a special sellerInfo product.

sttomm: ah, alright. then you could go with the separate sections or the central place for all public cloud platforms as proposed by me in the previous comment. it's up to you :)

sttomm (approved):

Jelledb (approved):

docs/meshstack.azure.metering.md
pubudug: I will go with this, because it's relevant only for the public clouds.

PR #358, opened 2021-05-26
feat: Improves Azure SPN setup docs

tfelix:
Updates it to the latest links from Azure and the latest API requirements.
Adds more information/insight about EA account setup.

CU-kd6xvb

Felix please read through it if you think something is misleading/incomplete. I tried to not directly describe every step but rather point to the respective Azure docs and only fill in the info misleading in Azure documentation, so in case they update something we are not having older information in our documentation. Or do you think this is a bad idea and should it rather be completly described by our own docs?


meshcloud-gmbh:
Task linked: [CU-kd6xvb Azure Integr. Docs not suitable for EA Account Setup](https://app.clickup.com/t/kd6xvb)

docs/meshstack.azure.index.md
```diff
@@ -54,136 +52,38 @@ It is important that the provided euid's are **case-sensitive** and must match t
...
-The meshcloud Azure [replication](./meshcloud.tenant.md) detects externally-provisioned subscriptions based on a configurable prefix in the subscription
-name. Upon assignment to a meshProject, the subscription is inflated with the right [Landing Zone](./meshstack.azure.landing-zones.md) configuration
-and removed from the subscription pool.
-
 ## Service Principal Setup
 
-meshStack uses separate service principals for different tasks following the best practice of least privilege principle.
-
-### Replicator
+meshStack uses separate service principals for different tasks following the best practice of least privilege principle. Depending on the way how you want to setup Subscription creation (there are two possible ways: Pre-provisioned Subscriptions or using an Enterprise Enrollment Account), you can use either an App Registration or an Enterprise Application. We recommend just using an Enterprise Application to keep things consistent and which is **required** to use the Enterprise Agreement Account based [Subscription provisioning method](#enterprise-enrollment-account).
```
felixzieger, line 57: q: I am still unsure about the relation App Registration <-> Enterprise Application. If I understood [this](https://www.seb8iaan.com/the-difference-between-azuread-app-registrations-and-enterprise-applications-explained/) correctly, an Enterprise Application is simply an instance of an App Registration. So the formulation here might be confusing to some Operators. 

Probably someone with more Azure experience like @SimonTheLeg or @younGihan could proof-read this specific section.

felixzieger:
In general, I think pointing to the Azure docs and explaining why we need them is fine.

But I am missing references to specific API versions outside of the REST API URIs you mention. If I understood it correctly, we need to give permissions via specific versions of some APIs. Are the specific versions only important for those APIs for which we specify the URIs? And what about the difference between legacy and new? I mean maybe we should only list the new version, once we are confident it works.

docs/meshstack.azure.index.md
```diff
@@ -36,10 +36,8 @@ In most organizations, other applications like Office 365 already consume user i
 
 ### Workload AAD Tenant
 
-Because meshcloud requires read-write permissions to manage user roles on Azure Subscriptions, we recommend creating a
-separate "Workload AAD Tenant" to be exclusively used by meshcloud. Our orchestration engine then creates Guest Users in the meshcloud AAD Tenant that reference user identities from the "home tenant" (AAD B2B). This way, users have a single cloud identity managed
-by your organization-wide policies while isolating "workload" related Azure activies into its own AAD Tenant which has no
-way of affecting other applications using the home tenant like Office 356 etc.
+Because meshcloud requires read-write permissions to your Active Directory to manage user roles on Azure Subscriptions, we recommend creating a
```
chewbeccy, line 39:
```suggestion
Because meshcloud requires read-write permissions to your Azure Active Directory to manage user roles on Azure Subscriptions, we recommend creating a
```

chewbeccy (approved):

docs/meshstack.azure.index.md
younGihan: Sounds right for the current context but there are also some other differences

younGihan (approved):

docs/meshstack.azure.index.md
felixzieger: Out of curiosity: What are other differences?

tfelix: hm yeah I wanted to keep the terminologic of the buttons you also find on the Azure portal panel. But maybe I should emphasis more that probably the best idea is to just use an Enterprise Application (which yea, is also an App Registration as well :D)

docs/meshstack.azure.index.md
tfelix: fixed

tfelix:
> But I am missing references to specific API versions outside of the REST API URIs you mention. If I understood it correctly, we need to give permissions via specific versions of some APIs. Are the specific versions only important for those APIs for which we specify the URIs? And what about the difference between legacy and new? I mean maybe we should only list the new version, once we are confident it works.

As far as I know Azure only mentioned the API difference for EA Account role assignment (the versions used to create an actual subscription and the role assignment obviously don't match, but I just used the version they mentioned). I try to make this a bit clearer.

I wanted to avoid describing the legacy way as this should not really be used anymore. So every new customer can just be onboarded via the new way. We btw need to change this for the existing customers as well!

felixzieger (approved):

PR #361, opened 2021-05-31
fix: update meshStack azure identity lookup snippet references

JohannesRudolph:


tfelix (approved):
yeah its a bit of a bummer that this update does not happen automactically. From time to time when doing a docs PR I try to remember to perform a snippet update, just in case.

PR #363, opened 2021-05-31
doc: consistently document private cloud metering supported resources

JohannesRudolph:


docs/meshstack.kubernetes.metering.md
```diff
@@ -13,4 +13,33 @@ meshStack supports metering and billing for most common Kubernetes versions.
...
+      status:
+        Status phase of this pod. See https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/
+    -}
+      { milliCpu : Integer, ramMb : Integer, status : Text }
+```
+<!--Example-->
+```dhall
+let example
+    : PodResourceTraits
+    = { milliCpu = +2500, ramMb = +1024, status = "Running" }
```
sttomm, line 43: d: is the + really supported here?

docs/meshstack.openshift.metering.md
```diff
@@ -13,4 +13,35 @@ meshStack supports metering and billing for most common OpenShift versions.
...
+```dhall
+let PodResourceTraits =
+    {-
+      milliCpu:
+        Configured total CPU limit of all containers in this Pod, in milli CPU (1/1000 of a CPU).
+
+      ramMb:
+        Configured total RAM limit of all containers in this Pod, in MiB.
+
+      status:
```
sttomm, line 36: d: We don't have a status predicate defined in our product catalog for pods. But as the status is available in teh traits, i guess it should work

docs/meshstack.cloudfoundry.metering.md
```diff
@@ -12,34 +12,76 @@ meshStack supports metering and billing for all versions of Cloud Foundry.
...
-traits:
- - SIZE (Instances * RAM)
- - EVENT
- - APP_STATE
-```
 
-### Marketplace
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Dhall Type-->
+```dhall
```
sttomm, line 23: d: I am wondering whether this dhall format is confusing, as the product catalog cannot be defined via dhall. And not sure whether it will in future. Putting the maintenance of the product catalog into the application is also a valid option as an alternative to dhall. That way admin could maintain the catalog by themself. But for sure, changing the catalog can be complex and may not be feasible for admins at the customer, but we'd have to think this through in the end which option fits better.

So for documenting it now like this, it is good to have the traits documented, but i am just not sure about the dhall format as it might bring readers into a wrong direction.

sttomm (changes requested):
i'll approve once we clarified about the format of how traits should be documented

docs/meshstack.kubernetes.metering.md
JohannesRudolph: that's unfortunately a bit of a dhall "oddity", i.e. a literal "2500" would be a `Natural` (floating point) while the "+2500" makes it an  `Integer`

docs/meshstack.openshift.metering.md
JohannesRudolph: While reviewing the trait definition I thought it was an oversight that it was not documented

docs/meshstack.cloudfoundry.metering.md
JohannesRudolph: I'm also not 100% sure if dhall is going to be the "primary" configuration format for the product catalog. But that's a decision I did not want to force at this time with this PR. My primary motivation was to get something (semi-)"automated" so we can generate the proper trait definitions and documentation of their values from meshStack's source code without (much) manual effort. The easiest path I saw to that goal was to reuse the existing dhall snippet infrastructure. 

Some additional thoughts: we will very likely also expose the product catalog via the API and in that case going from dhall -> YML -> meshObject API becomes very feasible, with the added advantage of having proper type definitions and ways to auto-generate catalogs with dhall. Of course we could also generate asciidocs like with the API... but that wasn't the obvious choice for me here either

sttomm: ok, sounds reasonable and especially the semi-automatic generation at the moment is a nice thing to have. So i am fine with that approach for now.

docs/meshstack.openshift.metering.md
sttomm: yeah, i am not sure which statuses are available for pods, but it makes sense to provide the option of filtering on status

docs/meshstack.kubernetes.metering.md
sttomm: ok

sttomm (approved):

PR #391, opened 2021-08-05
fix: add a remark about missing support for templates in Openshift 4

j0g3sc:


docs/meshstack.openshift.index.md
```diff
@@ -16,6 +16,7 @@ To enable integration with OpenShift, operators deploy and configure the meshSta
 ### OpenShift Versions
 
 meshStack currently supports OpenShift version 3.9+ as either Open-Source (OKD) or OpenShift Enterprise variants.
+The Openshift 4 integration currently does not support templates, so Landing Zones on Openshift 4 cannot make use of templates.
```
sttomm, line 19: d: Don't we also have a small section about applying templates via Landing Zones? If we do, you should mention the limitation there too.

sttomm (changes requested):

docs/meshstack.openshift.index.md
j0g3sc: done

henryde (approved):

tfelix (approved):

sttomm (approved):

PR #394, opened 2021-08-10
feat: document kubernetes pricing units

JohannesRudolph:


docs/meshstack.billing-configuration.md
```diff
@@ -94,6 +94,26 @@ let example2 =
...
+Traits are a properties of a cloud resource like CPU or RAM. Operators can use traits to configure predicates (filter resources applicable to a pricing rule) and to control how meshMetering calculates usages.
+
+meshMetering supports flexible rules for creating usages, like
+
+- **time**: charge consumption based on the usage of a resource over a time period
+- **quantity**: charge consumption based on a quantity
+- **time-quantity**: charge consumption based on the product of time and quantity
+
+Quantities are represented with prefixable units according to UCUM standard inside meshMetering.
+For example, one kilo byte would be represented as `1 kB` or one kibi-byte as `1 kiBy`.
```
henryde, line 111: q: Why is it not `1 kiB` even though it's byte in both cases?

henryde (approved):

docs/meshstack.billing-configuration.md
JohannesRudolph: We are using the UCUM standard https://ucum.org/ucum.html and §48 lists that for bytes its "By". (theres instead a mistake there in 1 kB). 

I will make this section more precise
- we use "print" units for formatting where units need to be human readable (e.g. in usage reports)
- we use the case-sensitive unit representations when units are processed automatically (traits, product catalog) and need to be serializable

sttomm (approved):

meshcloud-gmbh:
Task linked: [CU-rvumwj OpenShift CPU prices use a non-standard unit mCPU](https://app.clickup.com/t/rvumwj)

PR #410, opened 2021-09-02
Update meshstack.azure.landing-zones.md

felixzieger:


docs/meshstack.azure.landing-zones.md
```diff
@@ -180,4 +180,6 @@ In order to make an Azure Function only accessible via the replicator's Service
 
     ![Assign the Application Role to SP](assets/azure_function/sp-role.png)
 
+5. "accessTokenAcceptedVersion" : "2" must be set in the app registration manifests of the replicator service principal and the app registration manifest of the Azure Function.
```
tfelix, line 183: d: According to the discription here: https://docs.microsoft.com/en-us/azure/active-directory/develop/reference-app-manifest it somehow controls which endpoint is used. I am not sure if we maybe should make the text here a bit more optional. Like: "in case you see an authorization error when the replicator wants to invoke the function try this" as I am not sure if this should be done for every customer. Maybe this triggers other problems in their environments? (it at least works for some without this modification).

tfelix (changes requested):

docs/meshstack.azure.landing-zones.md
felixzieger: Good point. I made the step optional. I still think it should stay in the end of the steps, because that is where the error would appear.

tfelix (approved):

PR #411, opened 2021-09-06
Make it clear, that this is a "how to" section

felixzieger:


felixzieger:
I rewrote this section because I'd like to have the headlines match the how-to character of the content. See example request in https://app.clickup.com/t/1896w94

Jelledb (approved):
LGTM.

I saw the linked ticket but that still makes me wonder why we don't have a direct link from the tag UI to this docs page. That way, users can discover that we have these docs while using the UI.

PR #412, opened 2021-09-07
fix: email attribute mapping and some other missing dhall types

pubudug:


chewbeccy (approved):

docs/meshstack.identity-lookup.md
```diff
@@ -166,7 +168,34 @@ let exampleWithSchmaExtension
...
+      https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0#json-representation
+
+      Attention: This check is only performed on the first attempt when a user is added/invited
+      to a customer. If this check is configured after some users were initially added to a
+      customer their euid's are not udpated.
+
+        userAttributeName:
+            The AAD's user (custom) attribute.
+    -}
+      { userAttributeName : Text }
```
JohannesRudolph, line 194: t: I really appreciate when documentation comes with examples of "common real world" values because it helps illustrate how I'd actually use it. You can simply add one or multiple `let exampleX : AzureEuidUserAttribute = {...}`within the same source snippet block to include it in the docs

docs/meshstack.identity-lookup.md
```diff
@@ -204,6 +233,56 @@ let example
...
+let AzureUsernameAttributeUpdate =
+    {-
+      When adding/inviting a new user to a customer a custom attribute property from the users AAD
+      extension schema can be used to fill in his username.
+
+      Cannot be used together with an AzureEmailAttributeUpdate.
+
+        propertyToUse:
+            The name of the user property which is used as the username value. E.g. 'mailNickname'.
+        formatString:
```
JohannesRudolph, line 253: s: empty line between props

docs/meshstack.identity-lookup.md
```diff
@@ -204,6 +233,56 @@ let example
...
+            A printf() compatible string to replace/modify the found property with. For example to add a prefix to
```
JohannesRudolph, line 254: d: Java printf?

docs/meshstack.identity-lookup.md
```diff
@@ -166,7 +168,34 @@ let exampleWithSchmaExtension
 ```
 
-To use a non-standard attribute as
+To use a non-standard attribute as euid
```
JohannesRudolph, line 171: d: it would be helpful if this would say what the standard attribute is, e.g. when `emailAttributeUpdate` is a `None AzureEmailAttributeUpdate`, what is the default attribute used?

docs/meshstack.identity-lookup.md
```diff
@@ -204,6 +233,56 @@ let example
...
+
+        propertyToUse:
+            The name of the user property which is used as the email value. E.g. 'userPrincipalName'.
+
+    -}
+      { propertyToUse : Text }
+```
+<!--END_DOCUSAURUS_CODE_TABS-->
+
+> As mentioned above, both AzureEmailAttributeUpdate and AzureUsernameAttributeUpdate cannot be used at the same time. This is because  meshStack requires either the standard username (which is the userPrincipalName), or the standard email to be used as the user's username or email to uniquely identify a user already in meshStack and coming from AAD lookup.
```
JohannesRudolph, line 284: d: I'm not sure I can follow this section. 

The way I understand what's written there, my though process goes: 
- so I cannot configure AzureEmailAttributeUpdate and AzureUsernameAttributeUpdate at the same time - the dhall model could technically prevent this using a typing trick, but it doesn't 
- from a first glance, `AzureEmailAttributeUpdate` sounds like it would set meshUser's email address to a configurable value and `AzureUsernameAttributeUpdate` the meshUser's username. => So why can I not use them at the same time? 

I may be missing or misunderstanding something. Maybe trying to avoid passive voice (e.g. "to be used" -> "X uses Y") and specifically saying meshUser/AAD user instead of "user" can help clarify this section

JohannesRudolph (approved):

docs/meshstack.identity-lookup.md
pubudug: Fixed

docs/meshstack.identity-lookup.md
pubudug: Added some examples.

docs/meshstack.identity-lookup.md
pubudug: Updated to say Java String.format because we seem to be using that.

docs/meshstack.identity-lookup.md
pubudug: Added defaults where defaults were there. Removed the `non-standard` text for those that didn't have a default.

docs/meshstack.identity-lookup.md
pubudug: * The dhall model could technically support that, but there is one customer for whom we have to have both configured until we solve the duplicate users issue. And even after that, I am not sure how we would name the attribute name then. `usernameOrEmailAttributeUpdate`? And also to have Spring read it from Yaml we would have to introduce another yaml attribute as follows
`usernameOrEmailAttributeUpdate.username.propertyToUse` or
`usernameOrEmailAttributeUpdate.email.propertyToUse`
whereas the euid related attributes would be in the format
`euidUserAttributeUpdate.propertyToUse`
We could do that, but only after the duplicate users issue is fixed at the customer dev and prod environments.

* The reason why they cannot be used at the same time is because once we use a converted username and email, we have no way to identify that a user from AAD lookup is the same user that we have in the database. We could as a solution do this conversion at the lookup, but after some discussions we decided to keep it where it is at a post user creation step. I could give you more details in person if required.

Let me know if you have any objections though, because I will merge this before noon to get it into the release.

JohannesRudolph: > And also to have Spring read it from Yaml we would have to introduce another intermediate yaml attribute as follows
usernameOrEmailAttributeUpdate.username.propertyToUse or

iirc @henryde correctly that can be solved by transforming from the "mesh" dhall layer to the "service yml config" dhall layer. 

> Let me know if you have any objections though, because I will merge this before noon to get it into the release.

not in particular, it seems you gave the concerns I raised proper thought and since I'm not well versed in the actual details of the matter here I'll defer to your judgement about the best course forward. THe only opportunity I see is to add a section that descibres how user lookup matches "lookup users" existing meshUsers, similar to the way the documentation has a section that describes how meshStack replicator uses euid to match meshUsers to platform users

pubudug: I will check the possibility of adding a section as you mention. But I will go ahead and merge this one.

PR #413, opened 2021-09-08
fix: review comment fix on earlier PR

pubudug:
https://github.com/meshcloud/meshcloud-docs/pull/412/files#r704130203

### DIFF
--- a/docs/meshstack.identity-lookup.md
+++ b/docs/meshstack.identity-lookup.md
@@ -40,7 +40,7 @@ let example
 ```
 <!--END_DOCUSAURUS_CODE_TABS-->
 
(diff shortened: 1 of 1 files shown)


docs/meshstack.identity-lookup.md
```diff
@@ -40,6 +40,10 @@ let example
 ```
 
+Operators have the option of disabling inviting users that are not listed in the identity provider. They can do so by setting the `deny-assigning-other-users` configuration option to `true`.
```
chewbeccy, line 43:
```suggestion
Operators have the option of disabling the invitation of users that are not listed in the identity provider. They can do so by setting the `deny-assigning-other-users` configuration option to `true`.
```

chewbeccy (approved):
one suggestion

PR #414, opened 2021-09-09
Update meshstack.azure.index.md

felixzieger:


tfelix (approved):
I will have a look into the whole section. I mean it was also not so clear on what the parameters of the REST query need to be.

PR #429, opened 2021-11-01
feat: add section for user / group provisioning

j0g3sc:
CU-1p69at5

### DIFF
--- a/docs/meshstack.user-group-provisioning.md
+++ b/docs/meshstack.user-group-provisioning.md
@@ -1,0 +1,112 @@
+---
+id: meshstack.user-group-provisioning
+title: User and Group Provisioning
+---
+
+Provisioning users and groups to meshStack works based on the [SCIM 2.0](http://www.simplecloud.info/) standard.
+We provide an SCIM API that enables external systems such as an Azure Active Directory (AAD) to directly
+manage up-to-date user and group information within meshStack. `Provisioning` means changes about users and groups
+in the external system are automatically mirrored to the provisioned equivalents in meshStack. For example, when
+the members of a certain AAD group are changing and provisioning is set up, those changes will be reflected in the group within meshStack as well
+without additional effort. That way we can model the complete lifecycle of users and groups automatically.
+
+## Provisioning with SCIM
+
+SCIM clients need to address users and groups in order to manage the creation, updates and deletion. Because they cannot know
+user or group identifiers in advance they will attempt to find a user or group by a unique attribute:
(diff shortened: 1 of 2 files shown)


docs/meshstack.user-group-provisioning.md
```diff
@@ -0,0 +1,112 @@
...
+Provisioning users and groups to meshStack works based on the [SCIM 2.0](http://www.simplecloud.info/) standard.
+We provide an SCIM API that enables external systems such as an Azure Active Directory (AAD) to directly
+manage up-to-date user and group information within meshStack. `Provisioning` means changes about users and groups
+in the external system are automatically mirrored to the provisioned equivalents in meshStack. For example, when
+the members of a certain AAD group are changing and provisioning is set up, those changes will be reflected in the group within meshStack as well
+without additional effort. That way we can model the complete lifecycle of users and groups automatically.
+
+## Provisioning with SCIM
+
+SCIM clients need to address users and groups in order to manage the creation, updates and deletion. Because they cannot know
```
chewbeccy, line 15: we should clarify that SCIM client are the source like AAD. I

docs/meshstack.user-group-provisioning.md
```diff
@@ -0,0 +1,112 @@
...
+}
+```
+
+## SCIM API limitations
+
+The SCIM standard introduces comprehensive functionality to work with user and group resources.
+The implementation of meshStack's SCIM API does not support all features of SCIM. Here is a list of current limitations:
+
+- Deletion of users is not supported.
+- Filtering options are reduced:
```
chewbeccy, line 79: this has nothing to do with the scoping of the sync right? This is how the SCIM client can make changes to existing identities in meshStack?

chewbeccy (commented):
Some remarks

docs/meshstack.user-group-provisioning.md
```diff
@@ -0,0 +1,112 @@
...
+    ]
+}
+```
+
+## SCIM API limitations
+
+The SCIM standard introduces comprehensive functionality to work with user and group resources.
+The implementation of meshStack's SCIM API does not support all features of SCIM. Here is a list of current limitations:
+
+- Deletion of users is not supported.
```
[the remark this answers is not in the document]
j0g3sc: oh very good point, will clarify that.

docs/meshstack.user-group-provisioning.md
j0g3sc: clarified that, those are filter options and parameters that you can apply to search users and groups.

chewbeccy (approved):

PR #430, opened 2021-11-03
feat: Adds information about the Kubernetes configuration

tfelix:
I just realized, while writing and going through the Azure docs once more, I might have made a mistake while testing. Because in order to fetch the AKS credentals it seems the user need a special permission on the cluster resource itself. Which currently is not set via replication. I might was mislead because the user I used probably was an Admin/Owner on that resource and thus the fetching of creds just worked. Need to check that tomorrow and in that case we probably need one more replication step and I must adapt the docs.

But its quite finished beside that so I already open a draft PR.

### DIFF
--- a/docs/meshstack.customer-group-sync.md
+++ b/docs/meshstack.customer-group-sync.md
@@ -444,7 +444,10 @@ let example
         { username = "identityconnectorapi"
         , password =
             Secret.Native "EXTERNAL_IDENTITYCONNECTOR_MESH_API_PASSWORD"
-        , authorities = [ Authority.EXTERNAL_MESH_OBJECT_IMPORT ]
+        , authorities =
+          [ Authority.EXTERNAL_MESH_OBJECT_IMPORT
+          , Authority.CUSTOMEROWNER_ASSIGN
+          ]
         }
       }
 ```
--- a/docs/meshstack.kubernetes.config.md
+++ b/docs/meshstack.kubernetes.config.md
@@ -1,0 +1,120 @@
+---
+id: meshstack.kubernetes.config
+title: Configuration Reference
+---
+
+This section describes the configuration of a Kubernetes Platform Instance in the meshStack configuration model at `mesh.platforms`.
+
+For easier reference the following sections break down the configuration model in multiple parts. The union of these defines the full configuration model.
+
+## Core Configuration
+
+The core configuration is the basic setup for Kubernetes. I
+
+<!--snippet:mesh.platforms.kubernetes.core#type-->
+
+
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Dhall Type-->
+```dhall
+let KubernetesPlatformCoreConfiguration =
+    {-
+      platform:
+        The meshPlatform identifier
+
+      access-token:
+        Access token of the Kubernetes Service Principal. For more information on how to create this
+        principal with the correct access rights, please check the meshcloud Kubernetes documentation.
+
+      base-url:
+        The Base URL of the Kubernetes Cluster API.
+
+      disable-ssl-validation:
+        Disables SSL validation of the client. Advisable only for testing.
+
+      aks:
+        Add this part of the config if this is an AKS cluster.
+    -}
+      { platform : Text
+      , access-token : Secret
+      , base-url : Text
+      , disable-ssl-validation : Bool
+      , aks : Optional Aks
+      }
+```
+<!--Example-->
+```dhall
+let example
+    : KubernetesPlatformCoreConfiguration
+    = { platform = "kubernetes.mylocation"
+      , access-token = Secret.Native "ey*********"
+      , base-url =
+          "https://aks-meshcloud-ABCDEFGH.aaa.westeurope.azmk8s.io:443"
+      , disable-ssl-validation = False
+      , aks = None Aks
+      }
+```
+<!--END_DOCUSAURUS_CODE_TABS-->
+
+## AKS Configuration
+
+If this is an AKS cluster, please provide this part of the configuration as well.
+
+<!--snippet:mesh.platforms.kubernetes.aks#type-->
+
+
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Dhall Type-->
+```dhall
+let Aks =
+    {-
+      groupNamePattern:
+        Users need a cluster access permission in Azure in order to get credentials. This property defines
+        how the name of this AAD group is chosen.
+        It consists of a Java String.format() format string, receiving the following arguments:
+
+          1. meshCustomer identifier
+          2. meshProject identifier
+          3. meshProject ID (numeric)
+          4. role name suffix (configurable via Landing Zone)
+          5. platform identifier
+          6. location identifier.
+
+      aksSubscriptionId:
+        Azure Subscription ID that contains the AKS Cluster.
+
+      redirectUrl:
+        If users are invited (to get permissioned on the AKS cluster) into the AAD this is the invitation
+        redirect link they see in their invitation mail (same as in the Azure platform configuration).
+
+      sendAzureInvitationMail:
+        Flag if invitation mails are send at all.
+
+      servicePrincipal:
+        Service principal of the replicator that is used to setup the Azure user permissions for the AKS cluster
+        access.
+    -}
+      { groupNamePattern : Text
+      , aksSubscriptionId : Text
+      , redirectUrl : Text
+      , sendAzureInvitationMail : Bool
+      , servicePrincipal : ServicePrincipal
+      }
+```
+<!--Example-->
+```dhall
+let example
+    : Aks
+    = { groupNamePattern = "aks-%s.%s-%4\$s"
+      , aksSubscriptionId = "1234-1234-1234-1234"
+      , redirectUrl = "https://example.com"
+      , sendAzureInvitationMail = False
+      , servicePrincipal =
+        { aad-tenant = "example.onmicrosoft.com"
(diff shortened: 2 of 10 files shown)


docs/meshstack.kubernetes.index.md
```diff
@@ -3,16 +3,18 @@ id: meshstack.kubernetes.index
...
 Operators should be able to successfully integrate any Kubernetes distribution of their chosing as long as it has the
 required APIs and configuration options available.
+We also support Azure AAD authenticated users to get seamless access to the AKS (Azure Kubernetes) cluster.
 
 ## Integration Overview
 
 To enable integration with Kubernetes, operators deploy and configure the meshStack Kubernetes Module. Operators can configure one or multiple `PlatformInstance`s of `PlatformType` Kubernetes. This makes Kubernetes available to meshProjects like any other cloud platform in meshStack.
 
-meshStack automatically configures Kubernetes namespaces and RBAC permissions to integrate SSO with [meshStack Identity Federation](./meshstack.identity-federation.md).
+Depending on which Kubernetes distribution you want to use you will need to follow different steps. Below are the steps for the [meshStack Identity Federation](./meshstack.identity-federation.md) integration as well as for the Azure Kubernetes Services.
```
sttomm, line 17: d: Can you also link to the Azure Kubernetes Services section here?

docs/meshstack.kubernetes.index.md
```diff
@@ -92,6 +79,35 @@ rules:
...
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - appliedclusterresourcequotas
+  - clusterresourcequotas
```
sttomm, line 99: d: Do we really need the ability to create clusterresourcequotas? This sounds to me like cluster wide quotas that operators can set, but as they are not tenant specific, they are not set by meshStack. Or did i misunderstand these quotas?

docs/meshstack.kubernetes.index.md
```diff
@@ -195,7 +214,9 @@ Operators need to securely inject these access token into the configuration of t
 
 ### Custom meshProject Roles
 
-If you want to use custom roles to be mapped to your meshProject roles (and not just the pre-defined `admin`, `edit` and `view` roles) you need to make sure to also list these roles in the clusterrole binding section for the meshfed-service principle. It is not allowed for the service-principle to bind roles granting more rights then itself has, so the right to bind these roles must be explicitly given.
+> If you want to use custom roles to be mapped to your meshProject roles (and not just the pre-defined `admin`, `edit` and `view` roles) you need to make sure to also list these roles in the Cluster Role binding section for the meshfed-service principle.
```
sttomm, line 217:
```suggestion
> If you want to use custom roles to be mapped to your meshProject roles (and not just the pre-defined `admin`, `edit` and `view` roles) you need to make sure to also list these roles in the Cluster Role binding section for the meshfed-service principal.
```

docs/meshstack.kubernetes.index.md
```diff
@@ -195,7 +214,9 @@ Operators need to securely inject these access token into the configuration of t
...
+
+It is not allowed for the service-principle to bind roles granting more rights then itself has, so the right to bind these roles must be explicitly given.
```
sttomm, line 219:
```suggestion
It is not allowed for the service-principal to bind roles granting more rights then itself has, so the right to bind these roles must be explicitly given.
```

docs/meshstack.kubernetes.index.md
```diff
@@ -214,3 +235,35 @@ For example if you plan to use a role named `my-custom-role` please change the r
...
+
+meshStack automatically configures Kubernetes namespaces and RBAC permissions to integrate SSO with [meshStack Identity Federation](./meshstack.identity-federation.md).
+
+## Azure Kubernetes Services
+
+In order to use the AKS cluster properly with meshStack you need to enable the Azure AD integration. A good description on what needs to be done can be found in the [AKS-managed Azure Active Directory integration](https://docs.microsoft.com/en-us/azure/aks/managed-aad) document from Microsoft.
+
+You need the Azure CLI version 2.0.61 or later installed and configured. Run az --version to find the version. If you need to install or upgrade, see Install Azure CLI.
+
+> TODO If the user permissions steps are required for the credential download!
```
sttomm, line 263: f: just to be sure we don't forget it, i also put a comment here that this most likely needs to be added.

docs/meshstack.kubernetes.landing-zones.md
```diff
@@ -0,0 +1,17 @@
+---
+id: meshstack.kubernetes.landing-zones
+title: Landing Zones
+---
+
+The Kubernetes Landingzone can be used to setup shared Namespaces which are created during replication for each meshProject.
```
sttomm, line 6:
```suggestion
The Kubernetes Landing Zone can be used to setup shared namespaces which are created during replication for each meshProject.
```

i think we never write landingzone in one word

sttomm: but in general, what does that sentence mean? Shared namespaces sounds strange to me. Isn't it more like we create dedicated namespaces for meshProjects in a shared AKS cluster, but isolation between namespaces is still given?

docs/meshstack.onboarding.md
```diff
@@ -104,7 +104,7 @@ let UserRoleRequest =
           When the role request is for a new meshUser, set the email address (either entered by the inviting
           person or retrieved from identity lookup) as the new meshUser's euid.
 
-          The effect of this setting may be overridden, as IdentityLookup configuration takes precedence over
+          The effect of this setting may be overriden, as IdentityLookup configuration takes precendence over
```
sttomm, line 107: f: Hmm, somehow you reverted some textual fixes here. I guess someone may have changed the text only in the docs, but not in dhall. can you sync those fixes?

sttomm (changes requested):

docs/meshstack.customer-group-sync.md
```diff
@@ -444,7 +444,10 @@ let example
         { username = "identityconnectorapi"
         , password =
             Secret.Native "EXTERNAL_IDENTITYCONNECTOR_MESH_API_PASSWORD"
-        , authorities = [ Authority.EXTERNAL_MESH_OBJECT_IMPORT ]
+        , authorities =
+          [ Authority.EXTERNAL_MESH_OBJECT_IMPORT
+          , Authority.CUSTOMEROWNER_ASSIGN
```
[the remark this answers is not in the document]
tfelix: hm I just regenerated the snippets. That one was fixed automatically. So I guess its hopefully included everywhere now.

docs/meshstack.kubernetes.config.md
```diff
@@ -0,0 +1,73 @@
+---
+id: meshstack.kubernetes.config
+title: Configuration Reference
+---
+
+This section describes the configuration of an Kubernetes Platform Instance in the meshStack configuration model at `mesh.platforms`.
```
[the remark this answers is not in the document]
tfelix: fixed

docs/meshstack.kubernetes.index.md
tfelix: sure

docs/meshstack.kubernetes.index.md
tfelix: fixed

docs/meshstack.kubernetes.index.md
tfelix: fixed

docs/meshstack.kubernetes.index.md
tfelix: Yeah this one was re-worked.

docs/meshstack.kubernetes.index.md
tfelix: Actually I am not sure. This was taken over from OpenShift permission set, so I assumed they should be similiar if not the same (a few adaptions needed to be done because of the Groups and Users). And I tested it with this setup and it worked. The replication step uses `V1ResourceQuotaSpec`. So yeah sounds to me that this is a bit unneccessairy too. But I dont find the time now to re-check and rework this whole permission set but I can create a ticket to look into OpenShift and Kubernetes permissions set afterwards.

docs/meshstack.kubernetes.landing-zones.md
tfelix: Yeah true. I just needed an introduction sentence. It was rather bad. Changed it too: `
The Kubernetes Landing Zone can be used to configure how Namspaces are created inside a shared Kubernetes Cluster. `

docs/meshstack.onboarding.md
tfelix: fixed some typos in dhall file.

docs/meshstack.kubernetes.index.md
```diff
@@ -3,16 +3,18 @@ id: meshstack.kubernetes.index
 title: Integration
 ---
 
+
 meshStack supports project creation, configuration, user management and SSO for Kubernetes clusters.
 Our Kubernetes integration is distribution-independent and is purely based on "vanilla" upstream Kubernetes services.
 Operators should be able to successfully integrate any Kubernetes distribution of their chosing as long as it has the
 required APIs and configuration options available.
+We also support Azure AAD authenticated users to get seamless access to the AKS (Azure Kubernetes) cluster.
```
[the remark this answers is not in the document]
tfelix: I can try to emphasize more AKS but I guess a matrix is not a good idea because we sell stuff we don't support yet. And putting the non-supported platforms in the public docs is probably not a good selling point. :D

chewbeccy (approved):

PR #469, opened 2022-02-28
feat: removed multiple policy rule assignment

tobiasweiss93:
CU-1ydq6ew

### DIFF
--- a/docs/administration.mesh-policies.md
+++ b/docs/administration.mesh-policies.md
@@ -14,17 +14,14 @@ right on the policies screen. After clicking on it, there is a redirection to th
 details need to be filled out. To create the policy, you need to add the policy name in the form. Then you need to add
 both the meshObjects from the dropdown menu. For ease of use, we already restricted the meshObject combinations which
 do not make sense. You can choose only possible meshObject combinations. Once you select both meshObjects, the section
-for adding rules will appear where you can select the tag of selected meshObject. It is a global rule which will be
-applied everywhere. The created policy has been applied to every configuration which has selected meshObjects.
+for adding tags will appear where you can select the tag of selected meshObject. This restriction will be
+applied everywhere. The created meshPolicy has been applied to every configuration which has selected meshObjects.
 
 ### How it works
 
 Let's consider a scenario where you want to set a policy for the selection of a meshLandingZone. Therefore you configure
-a policy with a rule on the environment tags. Operator 'in' works like an intersection between the meshObjects. So if
-you want to assign a landing zone with a "prod" tag to a project which has a "prod" tag, it is valid. It is possible to
-assign multiple rules to a policy. But at least one rule has to be configured. In order to map two meshObjects
-successfully one of the policies defined for a match needs to be true. In order for a policy to return true, the rules
-have to be met.
+a policy on the environment tags. Operator 'in' works like an intersection between the meshObjects. So if
+you want to assign a landing zone with a "prod" tag to a project which has a "prod" tag, it is valid.
 
 ## Edit meshPolicies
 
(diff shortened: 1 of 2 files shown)


docs/meshcloud.policies.md
```diff
@@ -5,7 +5,7 @@ title: meshPolicies
 
 ## Introduction: what are meshPolicies?
 
-A meshPolicy is a set of rule(s) between two [meshObjects](meshcloud.index.md#introduction) that are globally defined by your organization. Each rule describes what [tag](meshcloud.metadata-tags.md) values have to be present on both meshObjects, in order to comply with the meshPolicy. meshPolicies are enforced in various places when attaching meshObjects. How meshPolicies can be configured is described [here](administration.mesh-policies.md).
+A meshPolicy describes a restriction between two [meshObjects](meshcloud.index.md#introduction) that are globally defined by your organization. Each meshPolicy describes what [tag](meshcloud.metadata-tags.md) values have to be present on both meshObjects, in order to comply with the meshPolicy. meshPolicies are enforced in various places when attaching meshObjects. How meshPolicies can be configured is described [here](administration.mesh-policies.md).
```
chewbeccy, line 8: I would really like the documentation and also the panel at some point reflect to two scopes of a policy. 1. the meshobject scope and 2. the tag scope. Instead of appear lets right "will be shown". Lets also write a sentence that the policy will protect the compliance of users, projects and environments in a multi-cloud setup. This is basically why we do this.

docs/administration.mesh-policies.md
```diff
@@ -14,17 +14,14 @@ right on the policies screen. After clicking on it, there is a redirection to th
...
 
 ### How it works
 
 Let's consider a scenario where you want to set a policy for the selection of a meshLandingZone. Therefore you configure
-a policy with a rule on the environment tags. Operator 'in' works like an intersection between the meshObjects. So if
-you want to assign a landing zone with a "prod" tag to a project which has a "prod" tag, it is valid. It is possible to
-assign multiple rules to a policy. But at least one rule has to be configured. In order to map two meshObjects
-successfully one of the policies defined for a match needs to be true. In order for a policy to return true, the rules
-have to be met.
+a policy on the environment tags. Operator 'in' works like an intersection between the meshObjects. So if
```
chewbeccy, line 23: can you please provide a bit more context around the example. First describing the policy and reason for the policy then what effects this will have in the product

chewbeccy (changes requested):
Bit more context for the example required. Also I'm missing information about having multiple policies for the same objects and how the behavior of the product will be

docs/administration.mesh-policies.md
tobiasweiss93: This is one of our first docs PRs. So I prefer that we don't add more information because we'll change the docs within the upcoming PRs. This means we would revert the docs again and again.
For now this change reflects the central change of the PR. We'll add more and more context information witht he upcoming changes.

docs/meshcloud.policies.md
sttomm, line 8:
```suggestion
A meshPolicy describes a restriction between two [meshObjects](meshcloud.index.md#introduction) that are globally defined by your organization. Each meshPolicy describes which [tag](meshcloud.metadata-tags.md) values have to be present on both meshObjects, in order to comply with the meshPolicy. meshPolicies are enforced in various places when attaching meshObjects. How meshPolicies can be configured is described [here](administration.mesh-policies.md).
```

sttomm (approved):

docs/meshcloud.policies.md
tobiasweiss93: We've discussed how we can improve the understanding. We'll add within an additional PR an diagram/image to improve the understanding.

tobiasweiss93: Fixed.

uleMeshcloud (approved):
lgtm

chewbeccy (approved):

PR #479, opened 2022-03-07
docs: prefer GCP SA authorization via Admin Console

JohannesRudolph:
Authorizing GCP Service Accounts for the Admin Groups API is easier to do with the Admin Console UI.
This used not to be available in the past, which is why we recommended setting it up via the Group API leveraging
the API explorer provided by Google.


malhussan (approved):
LGTM

PR #484, opened 2022-03-16
feat: Add missing documentation for replication operator role.

Jelledb:


felixzieger (approved):
The menu at meshcloud-dev also shows the 3 "customer admin/owner/employee" roles. Not sure if they should be shown on partner level at all. If yes, I'd say it makes sense to add them to the list as well.

image

Jelledb:
@felixzieger I have opened a PR in meshStack to get rid of these roles in the Partner area indefinitely. So I will merge this now.

PR #491, opened 2022-03-25
feat: Mention Payment Method Request button

Jelledb:


docs/meshcloud.payment-methods.md
```diff
@@ -88,3 +88,15 @@ The other way of exporting metadata is via our [chargeback statements](meshcloud
...
 
 The actual export itself can be done via the meshPanel. This is possible for both Partner Admins (for all meshCustomers) and Customer Admins (for the selected meshCustomer). When navigating to the Administration area (for partner admins), you will see **Chargeback Statements** on the left. When navigating to the [customer control plane](./meshcloud.customer.md#managing-your-meshcustomer) (for customer admins) , you will see **Chargeback Statements** under **Financials**. Click on it and you will see all chargeback statements. Additionally, there is the option at the top right labelled 'CSV Export' to export the list to a single CSV file.
+
+### External Payment Method Registration
+
+Beside manually creating a new payment method in the meshPanel on an ad-hoc basis, it is also possible to let users navigate to an external URL for requesting
+a new payment method, e.g. an internal budget request form. To make this even better, you can use the API of meshStack to create an approval flow and automate the creation
+of any new payment methods.
+
+The external URL can be figured under the configuration option `environment.ui.externalPaymentMethodUrl`.
```
tfelix, line 98: t: usually I think it would be better to have a complete dhall reference in here. But since the panel config (I believe) is not prepared for this that would be a ton of work now. But we should maybe expose the panel dhall config more into the documents.

tfelix (approved):

docs/meshcloud.payment-methods.md
Jelledb: Or make everything self-service 😈

PR #539, opened 2022-07-12
Controller role finance permissions

SimonBrockhaus:
Added chargeback statements to controller role in the overview table.


Jelledb (changes requested):
I noticed that this document is a bit chaotic and is lacking some information. It would be great if you could add that:

There is no mention of Payment Methods at all. I suggest we place them under 'Customer List'. Please create two rights: "Payment Methods List" and "Manage Payment Methods".

Additionally: we decided to add the Usage Report right. Let's add that one.

SimonBrockhaus:
image
image

Jelledb (approved):

PR #544, opened 2022-08-24
feat: adapt and extend single tenant deletion

uleMeshcloud:
* convert hisotry events and description of customer and project into a table
instead of
image
it's now
image
 
Before merging
- [x]  add release version

Since we have several places about project delete I hope I catch all of them

CU-arwc5j


tfelix (approved):
lgtm only minor remarks

uleMeshcloud:
mention that they loose access when meshTenant delete is triggered

Jelledb (changes requested):
Overall it looks good but I'm still confused about something:

The way I currently read it, it sounds as if projects automatically get deleted if all tenants are deleted. But this is not the case when you individually remove all tenants of a project, right? Just when you delete the project and trigger a batch deletion of tenants.

uleMeshcloud:
> Overall it looks good but I'm still confused about something:
> 
> The way I currently read it, it sounds as if projects automatically get deleted if all tenants are deleted. But this is not the case when you individually remove all tenants of a project, right? Just when you delete the project and trigger a batch deletion of tenants.

right. Maybe you can just point on the section where it makes you give that impression. I will try make it more clear.

docs/administration.projects.md
```diff
@@ -79,21 +77,26 @@ This export is an extension of the [Project Export](#project-export). It adds th
...
-- **EDIT_PROJECT_TAGS**: If Project Tags are configured, this event is written every time a tag is edited by a customer admin.
-- **PARTNER_EDIT_PROJECT_TAGS**: If Project Tags are configured, this event is written every time a partner/admin edits the restricted project tags.
-- **QUOTA_EDITED**: Quota Settings for the project were changed.
-- **SERVICE_USER_CREATED**: A Service User was created for the project.
-- **SERVICE_USER_DELETED**: A Service User was deleted for the project.
+| Event | Description|
+| ------| -----------|
+| **PROJECT_CREATED**| A new meshProject was created |
+| **PROJECT_PROJECT_MARKED_FOR_DELETION**<br>`since v7.x.x.`| A meshProject was deleted by a customer admin.|
+|**PROJECT_DELETED**| `since v7.x.x` When the deletion of all tenants is successful, this event is written.<br>`until v7.x.x`A meshProject was deleted by a customer admin.|
```
Jelledb, line 84: @uleMeshcloud here for example

docs/meshcloud.project.md
```diff
@@ -71,31 +71,22 @@ Via "Projects" -> "Expired Access", the expired or soon to expire role assignmen
...
+In the **Project Access** section you can click the `-` button in the row of a user or group to remove them from the project. The users and members will not be able to access this project in meshPortal and the cloud platforms anymore. You can add the user or group to your project again later on and all related users will get access again.
 
 ## Delete a meshProject
 
-If you would like to delete a project which is no longer used, open the corresponding project and navigate to **Settings** > **Danger Zone**. In this screen you can execute the project deletion.
+If you would like to delete a project which is no longer used, open the corresponding project and navigate to  **Danger Zone**. A meshProject is deleted when the deletion of all meshTenants have been performed successfully. In this screen you can execute the deletion of all meshTenants within your  meshProject. You can also [delete a single meshTenant](meshcloud.tenant.md#delete-a-meshtenant) on the tenant control plane.
```
Jelledb, line 78:
```suggestion
If you would like to delete a project that is no longer used, open the corresponding project, navigate to  **Danger Zone**, and start the deletion here. The project deletion will be finalized when the deletion of all meshTenants in the project has been performed successfully. If you want to keep the project but delete one or more individual tenants, you can also [delete a single meshTenant](meshcloud.tenant.md#delete-a-meshtenant) on the tenant control plane.
```

docs/meshcloud.project.md
```diff
@@ -71,31 +71,22 @@ Via "Projects" -> "Expired Access", the expired or soon to expire role assignmen
...
-
-**3. Combination of meshTenants**: If a projects contains a combination of tenants from 1. and 2. the tenants which don't require manual deletion are deleted automatically. For the other tenants the manual deletion step is necessary.
+> Your meshProject is not deleted when any meshTenant requiring a manual deletion has been declined.This means, your meshProject will be available on the customer control plane containing only the meshTenants which deletion had been declined.
```
Jelledb, line 92:
```suggestion
> Your meshProject is not deleted when any meshTenant requiring a manual deletion has been declined. This means, your meshProject will be available on the customer control plane containing only the meshTenants whose deletion had been declined.
```

docs/meshcloud.tenant.md
```diff
@@ -27,3 +27,35 @@ meshStack automatically derives [metadata tags](./meshcloud.metadata-tags.md) fo
 the meshCustomer it belongs to.
 
 Any update to tenant metadata (e.g. a change in payment method) triggers a new meshTenant reconciliation cycle.
+
+## Delete a meshTenant
+
+> Only users or users of [group](meshcloud.customer.md#user-groups) with the role [Customer Admin](meshcloud.customer.md#assign-meshcustomer-roles) have access to the administrative functionality described in this section.
```
Jelledb, line 33:
```suggestion
> Only users with the role [Customer Admin](meshcloud.customer.md#assign-meshcustomer-roles) or [Customer Owner](meshcloud.customer.md#assign-meshcustomer-roles) have access to the administrative functionality described in this section.
```

Jelledb: I know groups would also work but I'd just keep it simple by mentioning 'users'.

docs/administration.projects.md
```diff
@@ -28,42 +28,40 @@ To edit the quotas, please follow these steps:
 5. Also under the **Actions** column, you reach the **Quota Dashboard**. These quotas are project-specific.
 6. You can now change the quotas as desired.
 
-### Delete Projects
+### Delete Tenants
```
Jelledb, line 31: q: Is this whole section maybe better fit for its own page? As it will also be its own page in the Admin Area. It feels a bit weird to me to place this in the projects page.

Jelledb (changes requested):
Still have some open topics that could be improved but it really is already a big improvement from the current documentation around this area!

docs/administration.projects.md
uleMeshcloud: ah thanks. I will adapt it to make it more clear

docs/administration.projects.md
uleMeshcloud: This page includes all functionalities regarding projects. So I wouldn't place it completely into a new page currently. Since we also don't do it for other models. Maybe we can take it up when having the overhaul of docs.

docs/administration.projects.md
```diff
@@ -79,21 +77,26 @@ This export is an extension of the [Project Export](#project-export). It adds th
...
-- **PROJECT_ROLE_REJECTED**: If [4-EP](meshstack.authorization.md#user-project-role-approval) is active, this event is written when a role request is rejected.
-- **EDIT_PROJECT_TAGS**: If Project Tags are configured, this event is written every time a tag is edited by a customer admin.
-- **PARTNER_EDIT_PROJECT_TAGS**: If Project Tags are configured, this event is written every time a partner/admin edits the restricted project tags.
-- **QUOTA_EDITED**: Quota Settings for the project were changed.
-- **SERVICE_USER_CREATED**: A Service User was created for the project.
-- **SERVICE_USER_DELETED**: A Service User was deleted for the project.
+| Event | Description|
+| ------| -----------|
+| **PROJECT_CREATED**| A new meshProject was created |
+| **PROJECT_PROJECT_MARKED_FOR_DELETION**<br>`since v7.x.x.`| A meshProject was deleted by a customer admin.|
```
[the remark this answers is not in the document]
uleMeshcloud: clarified f2f:
since label on the left side = is a new Event
since label on the right side = indicates when the new meaning applies

Jelledb (approved):

PR #545, opened 2022-08-24
feat: Describe budget alert functionality.

Jelledb:


website/sidebars.json
```diff
@@ -12,6 +12,7 @@
       "meshcloud.project-resources",
       "meshcloud.project-metering",
       "meshcloud.payment-methods",
+      "meshcloud.budget-alerts",
```
tfelix, line 15: d: question would be if we want to actually add a new entry for every alert of if we want to group this under a more generic "alerts" section. But I guess this question can be re-asked when multiple alerts descriptions are added.

tfelix (approved):
lgtm

website/sidebars.json
Jelledb: Good point. I do think in general that this one is special enough to mention on its own.

PR #548, opened 2022-09-07
fix: Reference api including index.html in path

sttomm:
without the index.html relative links inside the docs don't work


tobiasweiss93 (approved):
Lgtm.

PR #586, opened 2022-11-24
Feature/update snippets

felixzieger:


docs/meshstack.customer-group-sync.md
```diff
@@ -169,6 +170,44 @@ let example =
...
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Dhall Type-->
+```dhall
+let IfNullAttributeTransformation =
+    {-
+      A IfNullAttributeTransformation takes a nullable attribute. If the value is null it uses a replacement value.
+      In a second step it applies an optional postProcessor on the value.
+
+      attribute:
+          The LDAP attribute that should be processed. Can be null. For example "mail"
```
tfelix, line 187: d: enabling null is a bit funny. I guess this would make this a nop in directly fallback to the ifNullAttribute?

tfelix (approved):

docs/meshstack.customer-group-sync.md
felixzieger: Thanks for the catch. That was formulated badly. The config value `null` does not have a special meaning, the attribute can be un-defined for an LDAP entity (e.g. an Service Account does not have a `mail` attribute). I updated the description.

PR #587, opened 2022-11-25
Update meshstack.replication-configuration.md

younGihan:
platform place holder actually provides the platformIdentifier not the Platform display name.


younGihan:
A Customer pointed out that it is confusing as we usually document if something is based on an identifier

tfelix (approved):

PR #605, opened 2023-02-01
docs: fix instruction text around navigation to the marketplace

uleMeshcloud:
fix instruction text around navigation to the marketplace as we notice in our workshop that it is not up to date anymore
Before:
image

After:
image

### DIFF
--- a/docs/meshstack.meshmarketplace.development.md
+++ b/docs/meshstack.meshmarketplace.development.md
@@ -3,21 +3,26 @@ id: meshstack.meshmarketplace.development
 title: Marketplace Development
 ---
 
-You can provide your own services in the [meshMarketplace](marketplace.index.md) via the Marketplace Development. The functionality is currently
-only available for [Customer Admins](meshcloud.customer.md#manage-groups-of-assigned-users).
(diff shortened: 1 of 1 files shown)


docs/meshstack.meshmarketplace.development.md
```diff
@@ -3,21 +3,26 @@ id: meshstack.meshmarketplace.development
...
+> Only users with the role [Customer Admin](meshcloud.customer.md#assign-meshcustomer-roles) or [Customer Owner](meshcloud.customer.md#assign-meshcustomer-roles) have access to the administrative functionality described in this section.
+
+You can provide your own services (e.g. databases, message brokers, filesystems, etc) in the [meshMarketplace](marketplace.index.md) via the **Marketplace** >  **Service Brokers** on your [customer control plane](./meshcloud.customer.md#managing-your-meshcustomer). This requires you have a running Service Broker, an application that manages these services by the means of [Open Service Broker API](https://www.openservicebrokerapi.org/). Services provided by you can then be consumed by other users of the meshPortal. A short overview and some specifics that should be considered when writing a Service Broker for the meshMarketplace are described [here](meshstack.meshmarketplace.index.md).
```
j0g3sc, line 8: q: is meshPortal a thing? I see that it was there before your changes but always thought its the meshPanel.

j0g3sc (approved):

docs/meshstack.meshmarketplace.development.md
uleMeshcloud: Yeah you're right. Good point. it seems to be an outdated term. It should be actually meshStack, as it is the users of meshStack who can consume it. But currently only via the panel. I will update it :+1:

PR #606, opened 2023-02-16
Feature/cu 861macwqu update public documentation

tfelix:
I found that the user & group section naming was not really in sync, thats why adapted it. Not sure if this is a good think. If you dont think I can revert it back.


docs/meshstack.identity-federation.md
```diff
@@ -365,3 +365,15 @@ let AzureLookupStrategy =
...
 ```
+
+## Group Cleanup
+
+For the platform listed below user access is given by assigning groups with the apropriate users inside to the platform tenant. The naming of these groups can be customized by setting the `User Group Name Pattern` in the appropriate meshPlatform Configuration. See the image below.
+
+![User Group Name pattern in the meshPlatform config](assets/platform-config-groupname-pattern.png)
+
+Additionally added group permissions are discouraged, but possible. During replication meshStack won't remove additionally added group permissions that external systems might have added.
```
kirilloak, line 375: Recently I found that at least for Azure, we removing all customer-added permission for groups, and trying to restore the desired state from our role mappings. Not sure about the other platforms

docs/meshstack.identity-federation.md
```diff
@@ -365,3 +365,15 @@ let AzureLookupStrategy =
...
+
+When the name of a group has changed by a user initiated config change e.g. when modifying the group name pattern inside the platform config or by changing the meshRole to Platform Role Mapping inside a Landing Zone as long as the parameter `#{platformGroupAlias}` is used in the User Group Name Pattern, the group is deleted and re-created with a different name.
+The deletion is done because certain cloud platforms don't support group renaming
+If additional tenant access permissions were attached to this group by an external system, these permissions are lost and not carried over.
```
kirilloak, line 379:
```suggestion
If additional tenant access permissions were attached to this group by an external system, these permissions are lost and not carried over.

Automatic group cleanup is available for GCP, AWS, Azure, Azure Kubernetes Service, and OpenShift.
```

kirilloak (approved):
Lgtm
P.S. I think it will be helpful to list platforms where group cleanup is available

docs/meshstack.identity-federation.md
tfelix: changed

docs/meshstack.identity-federation.md
tfelix: I adapted it to reflect our latest findings.

PR #608, opened 2023-03-06
chore: remove obsolete info about payment method type

nroi:
Removed the fifth entry that describes to select the payment method type.

As discussed in slack:

> In the past we had this as a direct field on the payment method. But it doe snot exist anymore. So we just have to remove it from the public docs.


PolinaSadykova (approved):

PolinaSadykova:
Sorry for the waiting time, completely forgot about it:(

PR #616, opened 2023-04-28
SCIM setup wording

felixzieger:


nroi (approved):
lgtm, except for the one remark.

Thanks for incorporating customer feedback into our docs!

PR #624, opened 2023-06-13
chore: remove unused markdown page

nroi:
I've stumbled on this file because the `id` and the `title` in the first two rows don't match with the file name. I wanted to fix it, but it seems this file isn't used anywhere. A simple `grep` for the file name returns no meaningful results.

Whoever approves this, please approve only if you understand how our website is built. I want the approval to make sure I haven't missed anything and to be sure this file is really not used anywhere.


Jelledb (changes requested):
You are correct that is not used right now, but it is a useful guide that we should add. However, it is a bit outdated. For it to fit our current state of the product again we need to update this guide from "How to confirm project deletion" to "how to confirm tenant deletion", which is the new way of doing things.

nroi:
Thanks for the info. I'm not completely in the loop with our recent changes regarding project deletion vs tenant deletion, so this is not the kind of change that I can do in just a few minutes. I'm closing this in favor of a flock log ticket: https://app.clickup.com/t/863h1w59c

PR #625, opened 2023-06-13
chore: add a nix file for working with meshcloud-docs

JohannesRudolph:
@henryde not sure if this is a good idea as it essentially copies a bunch from our meshStack repo nix file and needs to be kept in sync then down the road, but i found it helpful

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,56 @@
+{ pkgs ? import <nixpkgs> { } }:
+
+# Fixing packages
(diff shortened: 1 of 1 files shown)


henryde:
> @henryde not sure if this is a good idea as it essentially copies a bunch from our meshStack repo nix file and needs to be kept in sync then down the road, but i found it helpful

Bettern than not having it, it also documents the dhall version dependency.

henryde (approved):

PR #673, opened 2023-09-28
feat: added SCIM only user management docs

tobiasweiss93:
CU-86bvz67


docs/meshstack.user-group-provisioning.md
```diff
@@ -181,3 +181,7 @@ displayed: "Create", "Update" and "Delete". Ensure that all three checkboxes are
 Deleting a user via SCIM has the following consequences in meshStack: The user will be irrevocably deleted and only the
 bare minimum of information required for GDPR compliance is retained. After deletion, it is possible to create a new
 user with the same username and email as the previously deleted user, but the deleted user cannot be restored.
+
+## SCIM only user management
+
+The default behavior of meshStack is that users can be created via panel, SCIM or any other way. In the admin area you can configure SCIM to be the single source of truth to avoid conflicts when the same user exists both in the IdP and in SCIM. You can find this option in Settings/Management.
```
PolinaSadykova, line 187:
```suggestion

## SCIM only option for User Lifecycle

By default, in meshStack, you can create users through a diverse range of methods like Panel, IDP, SCIM, and other available options. Now you can choose an option to manage users and groups within the Workspace and Admin areas exclusively via SCIM to prevent conflicts and double assignments when other methods are available. To enable creation only via SCIM go to the Settings page on the Admin area and in the Manage tab switch on a "SCIM only user management" button. The button is switched off by default, which means that users and groups can be created through a diverse range of methods.
```

PolinaSadykova (approved):
Thanks a lot! I have updated it a bit :) Can we also add a screenshot of UI, which you created?

PolinaSadykova:
image

docs/meshstack.user-group-provisioning.md
tobiasweiss93: Fixed.

tobiasweiss93:
> image

Fixed

PR #680, opened 2023-10-17
Adding to user docs information about Admin Groups

PolinaSadykova:
changed user docs based on Metro feedback about Admin Groups


obinna-okoro (approved):
lgtm

PolinaSadykova:
@obinna-okoro Thanks, I fixed it :)

PR #682, opened 2023-10-20
fix: Algolia search works again

Jelledb:
* I am not sure how this worked initially but I found the two indexes on Algolia and pointed to their App IDs & API Key
* I changed the code in the Crawler so it correctly picks up headers (https://github.com/facebook/docusaurus/issues/6432)
* I changed in general some of the URLs that were pointed to in the Crawler.

All in all this seems to work now. You can run `yarn build:master` locally and serve the build and it will work :-)

Demo: https://jmp.sh/erIJKzE9

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -6,9 +6,9 @@
   "scripts": {
     "examples": "docusaurus-examples",
     "start": "docusaurus-start",
-    "build:master": "DOCSEARCH_INDEXNAME='meshcloud' DOCSEARCH_APIKEY='1b6387d9fcda3cba4cdb3ffb68cde260' docusaurus-build",
+    "build:master": "DOCSEARCH_INDEXNAME='meshcloud' DOCSEARCH_APIKEY='aa3b874dff5c832fe2e3ed42a8062160' DOCSEARCH_APP_ID='LDDGX81P02' docusaurus-build",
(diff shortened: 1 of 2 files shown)


website/package.json
```diff
@@ -6,9 +6,9 @@
   "scripts": {
     "examples": "docusaurus-examples",
     "start": "docusaurus-start",
-    "build:master": "DOCSEARCH_INDEXNAME='meshcloud' DOCSEARCH_APIKEY='1b6387d9fcda3cba4cdb3ffb68cde260' docusaurus-build",
+    "build:master": "DOCSEARCH_INDEXNAME='meshcloud' DOCSEARCH_APIKEY='aa3b874dff5c832fe2e3ed42a8062160' DOCSEARCH_APP_ID='LDDGX81P02' docusaurus-build",
```
tfelix, line 9: d: I hope this APIKEY is not sensitive so its okay to have it in the public repo? Not that someone can extract this key and post a own, modified search result to the service? Not sure how it works. I thought during build time it generates an index and transmits to their server? Then to have those keys here I guess could be problematic?

tfelix (approved):
In general okay (as the key was there before anyways) but I added a remark and question if this is actually a good idea.

website/package.json
Jelledb: image

The key is just for querying in frontend and is safe.

PR #693, opened 2023-11-17
feat: rename customer to workspace in deleted tenants page and landin…

uleMeshcloud:
…g zone

Actually, I want to test some integration but I thought it would be useful to push some meaningful changes along the way if I need to make changes in order to test things.
Will update the content of the ticket, as it linked to a meaningless Delete me ticket
CU-86bwfhe88

### DIFF
--- a/docs/administration.delete-tenants.md
+++ b/docs/administration.delete-tenants.md
@@ -27,4 +27,4 @@ In order to confirm the deletion, you have to first perform the manual deletion 
 
 If you choose to decline the deletion, you can do so by clicking on the decline button. If you decline the deletion, the meshTenant will be available again on the workspace control plane.
 
-> If a meshProject was marked for deletion before by a customer, and all meshTenants of the meshProject have been successfully deleted, the meshProject is permanently deleted.
+> If a meshProject was marked for deletion before by a workspace user, and all meshTenants of the meshProject have been successfully deleted, the meshProject is permanently deleted.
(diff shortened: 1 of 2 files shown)


docs/administration.delete-tenants.md
```diff
@@ -27,4 +27,4 @@ In order to confirm the deletion, you have to first perform the manual deletion
 
 If you choose to decline the deletion, you can do so by clicking on the decline button. If you decline the deletion, the meshTenant will be available again on the workspace control plane.
 
-> If a meshProject was marked for deletion before by a customer, and all meshTenants of the meshProject have been successfully deleted, the meshProject is permanently deleted.
+> If a meshProject was marked for deletion before by a workspace, and all meshTenants of the meshProject have been successfully deleted, the meshProject is permanently deleted.
```
nroi, line 30: d: afaik we've used the term `customer` before to refer to either the customer construct within meshStack, or to an actual person associated to a `customer`. But this doesn't work anymore with the new name, because a workspace is not a person, so I would just use the term `workspace user`.

```suggestion
> If a meshProject was marked for deletion before by a workspace user, and all meshTenants of the meshProject have been successfully deleted, the meshProject is permanently deleted.
```

nroi (approved):
thanks!

docs/administration.delete-tenants.md
uleMeshcloud: good catch. Mindless search & replace does not work everytime :'D

PR #700, opened 2023-12-04
fix: Updates SCIM related text

tfelix:
Explains the now mandatory SCIM flag.


docs/meshstack.user-group-provisioning.md
```diff
@@ -10,6 +10,16 @@ in the external system are automatically mirrored to the provisioned equivalents
 the members of a certain AAD group are changing and provisioning is set up, those changes will be reflected in the group within meshStack as well
 without additional effort. That way we can model the complete lifecycle of users and groups automatically.
 
+## Enable SCIM
+
+By default, in meshStack, you can create users through a diverse range of methods like the Panel, IDP, and other available
+options. If you want to manage your users via SCIM exclusively you need to enable this setting in the Admin area
+(this disabled all other methods to prevent conflicts and double assignments).
+To enable SCIM go to the Settings page on the Admin area and in the Manage tab switch on a "SCIM only user management" button.
+The button is switched  off by default, which means that users and groups can be created through a diverse range of methods.
```
chewbeccy, line 19:
```suggestion
The button is switched off by default, which means that users and groups can be created through a diverse range of methods.
```

docs/meshstack.user-group-provisioning.md
```diff
@@ -10,6 +10,16 @@ in the external system are automatically mirrored to the provisioned equivalents
 the members of a certain AAD group are changing and provisioning is set up, those changes will be reflected in the group within meshStack as well
 without additional effort. That way we can model the complete lifecycle of users and groups automatically.
 
+## Enable SCIM
+
+By default, in meshStack, you can create users through a diverse range of methods like the Panel, IDP, and other available
```
chewbeccy, line 15: By default, in meshStack, users can be created using a variety of methods such as the Panel, IDP, and other available options. If you prefer to manage your users exclusively via SCIM, you need to enable this setting in the Admin area. This disables all other methods to prevent conflicts and double assignments. To enable SCIM, go to the Settings page in the Admin area, and in the 'Manage' tab, turn on the 'SCIM only user management' button. By default, this button is turned off, allowing users and groups to be created through other means."

felixzieger (approved):
I added a step in the AAD instructions

PolinaSadykova:
@tfelix we also have here about SCIM Settings kind of the same info  https://docs.meshcloud.io/docs/meshstack.user-group-provisioning.html Should we then maybe delete this info?

chewbeccy (approved):

docs/meshstack.user-group-provisioning.md
tfelix: Switched it to your text and also updated the screenshot.

tfelix:
@PolinaSadykova no this one is still relevant as it describes the SCIM integration which has not changed. Its only the meaning of this setting that has changed.

docs/meshstack.user-group-provisioning.md
tfelix: f

PR #724, opened 2024-01-08
feat: Describe custom platform capability

Jelledb:


radu-ux-meshie (approved):
I think overall this is fine, but I have the feeling we can improve it: make it shorter and easier to read.

PR #735, opened 2024-01-24
Docs for one catalog CR

PolinaSadykova:
- Adding 2 pages with guides (onbaording & adding platform services) ->checked with @chewbeccy 
- general info about marketplace -> checked with @chewbeccy 
- Rename meshMarketplace to OSB Services and change structure of docs around old marketplace


uleMeshcloud:
closed cleaned content merged on dev already

PR #749, opened 2024-02-14
feat: Mention deprecate/undeprecate use case for OSBs

Jelledb:


docs/meshstack.meshmarketplace.index.md
```diff
@@ -49,9 +49,24 @@ OSB Services platform supports two different options for authenticating and auth
...
 
-## Platform Instance Configuration
+## Deprecating OSB Services
 
-OSB Services platform can only be configured by meshcloud via Dhall. It is not yet available in self-service. Please consult the following example as a reference of possible configuration settings.
+At some point along the lifecycle of an OSB service you might want to deprecate it, i.e. existing service instances will
+remain and keep working but you prevent the booking of new service instances.
+
+You can do so by removing the service definition from the service catalog. meshStack will retain all known information about
+the service and service instances but from there on it will no longer be possible to see the service and book new instances of
```
sttomm, line 58:
```suggestion
the service and service instances but from there on it will no longer be possible to book new instances of
```
I think the "see" part might confuse, that you also can't see the service instances anymore. But you actually still can. It is just not available in the marketplace(s) anymore

sttomm (approved):

PR #779, opened 2024-04-03
dns and ssl

felixzieger:
image

@henryde @malhussan Did I forget an endpoint in that list?


malhussan (approved):
It might also make sense to point that this is mainly for Enterprise, for SaaS we usually go with the CNAME approach but manage certificates ourselves

felixzieger:
I decided to not mention the rvn-0 domain, since this is also not documented anywhere else in the docs yet and is needed in rare cases. If this becomes more frequent, we extend the list here

PR #789, opened 2024-04-23
feat: Describe tenant deletion with new flows

Jelledb:


JohannesRudolph:
lgtm, i extended the docs slightly to describe the process end to end and added the required permissions for setting up replicator

PR #808, opened 2024-05-22
Feature/identity and access

felixzieger:


docs/meshstack.user-group-provisioning.md
```diff
@@ -143,6 +143,33 @@ To set up the provisioning on AAD side, have a look at [Microsoft's guideline](h
 | Join(" ", [givenName], [surname])                             | name.formatted                             |
 | Attribute which is used as euid in your meshStack             | externalId                                 |
 
+### How to map attributes via expressions
+
+A common patttern we see is that Admin Users do not have email attributes filled. Here is how to create a mapping that sets the email attribute based on an expression.
```
JohannesRudolph, line 148: Can you add a quick intro for what we mean with "admin users"? Like

"Some organizations requires the use of dedicated user accounts for administration purposes (segregation of duties). Unlike "normal users" used for everyday office work, these "admin users" often  do not have email attributes" etc.

docs/meshstack.user-group-provisioning.md
```diff
@@ -143,6 +143,33 @@ To set up the provisioning on AAD side, have a look at [Microsoft's guideline](h
 | Join(" ", [givenName], [surname])                             | name.formatted                             |
 | Attribute which is used as euid in your meshStack             | externalId                                 |
 
+### How to map attributes via expressions
```
JohannesRudolph, line 146: In many situations it can be useful to modify user attributes before syncing them to meshStack. For example...

website/sidebars.json
```diff
@@ -75,7 +75,6 @@
       "meshstack.identity-provider",
       "meshstack.identity-lookup",
       "meshstack.authorization",
-      "meshstack.user-revocation",
```
JohannesRudolph, line 78: perfect, that page is indeed outdated and no longer required!

JohannesRudolph (approved):

docs/meshstack.user-group-provisioning.md
felixzieger: Good idea, I added a motivation to the how

PR #855, opened 2024-07-22
chore: Delete Project Resources

Jelledb:


chewbeccy (changes requested):
Please provide links where the information can be found instead before we delete those sections

chewbeccy (approved):

PR #859, opened 2024-07-22
chore: Delete Resource Inventory page

Jelledb:


Jelledb:
I will close this again for now as we are still deciding this

PR #918, opened 2024-09-26
fix: Wrong favicon color

Jelledb:
Reynold made me aware that for the favicon (the icon in the browser tab) is the wrong color. It should have blue on a white background.


uleMeshcloud (approved):
lgtm

PR #980, opened 2025-01-10
feat: add info about when to process chargebacks

nroi:
CU-86c15vwp3

I've renamed "cloud foundation team'" to "platform team" in some cases, here's the related discussion: https://meshcloud.slack.com/archives/CSLGLMN73/p1736507492578749

Also note that I didn't explicitly describe the behavior that it can happen that the chargeback is first finalized, and then updated by a new chargeback (due to our cleanup job), because I was not sure how to describe this both concisely and correctly. And imho, a short text that gives the customer clear instructions on what to do is more valuable than a huge wall of text that is hard to understand.

But if you have a suggestion on how to describe this without overwhelming the customer with too many details, I can add it.

### DIFF
--- a/docs/meshcloud.project-metering.md
+++ b/docs/meshcloud.project-metering.md
@@ -75,6 +75,13 @@ This export will contain the line items (see above) of all the chargeback statem
 The line item data is suitable for feeding into chargeback processing, e.g. importing it to an ERP System to transfer
 budgets between cost centers.
 
+For customers that export chargeback statements on a monthly basis, the following recommendation applies:
+Chargeback statements of the previous month should be processed not earlier than the `finalizeReportsAfterDays` setting,
+plus 2 days, at 7 am (German local time). For example, if `finalizeReportsAfterDays` is set to 4
+(which is the default), then chargeback statements of May can be processed starting at June 6th, 7am.
+If you don't know the configuration of the `finalizeReportsAfterDays` setting in your meshStack, please contact your
+platform team.
+
 Chargeback Statements also contain billing information per line item. Your platform team can 
 [configure](meshstack.billing-configuration.md)
 which information meshStack should include as billing information in chargeback statements.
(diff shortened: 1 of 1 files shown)


JohannesRudolph:
Task linked: [CU-86c15vwp3 Fix finalization of partly incomplete chargeback statements](https://app.clickup.com/t/86c15vwp3)

docs/meshcloud.project-metering.md
```diff
@@ -75,13 +75,22 @@ This export will contain the line items (see above) of all the chargeback statem
...
-> Cloud Foundation teams typically configure billing information to payment method name, identifier, expiration date and amount as well as any workspace tags, project tags and payment method tags.
+> Platform teams typically configure billing information to payment method name, identifier, 
+> expiration date and amount as well as any workspace tags, project tags and payment method tags.
 
 You can review this billing information in meshPanel when opening chargeback statement. CSV Exports of chargeback statements also include the configured billing information.
 
+For customers that export chargeback statements on a monthly basis, the following recommendation applies:
+Chargeback statements of the previous month should be processed not earlier than the `finalizeReportsAfterDays` setting,
+plus 2 days, at 7 am (German local time). For example, if `finalizeReportsAfterDays` is set to 4 
+(which is the default), then chargeback statements of May can be processed starting at June 6th, 7am.
```
nroi, line 90: regarding the 7 am: the cron job actually runs at 04:33, but: this is the system time zone, which is probably UTC, so this will translate to 05:33 or 06:33 (depending on whether we have daylight savings time or not). We don't want to annoy our customers with such technical details, so we just say 07:00, which should be safe regardless of whether we have daylight savings time or not.

docs/meshcloud.project-metering.md
```diff
@@ -75,10 +75,19 @@ This export will contain the line items (see above) of all the chargeback statem
 The line item data is suitable for feeding into chargeback processing, e.g. importing it to an ERP System to transfer
 budgets between cost centers.
 
-Chargeback Statements also contain billing information per line item. Your Cloud Foundation team can [configure](meshstack.billing-configuration.md)
+For customers who export chargeback statements on a monthly basis, the following recommendation applies:
+Chargeback statements of the previous month should be processed no earlier than the `finalizeReportsAfterDays` setting
```
sttomm, line 79: d: sounds good to me. I just think we also should have it at another place. This one here is about CSV export and it makes sense to mention it here. But the other way and the preferred way to get the CBs is via the API. So we should either mention it additionally at a more central place, if it exists in these docs, or we could add it to the API docs of the meshChargeback API.

sttomm (changes requested):
lgtm, I just think we have to add the same info somewhere else too. Therefore requesting changes, but may just approve if the other place is the API docs

sttomm (approved):
lgtm

PolinaSadykova (approved):

PR #987, opened 2025-01-20
fix: clarify docs around chargeback v2

nroi:


sttomm (approved):
lgtm, but Polinas should have a final look

PolinaSadykova (approved):
thanks!

PR #1049, opened 2025-04-04
feat: Document how to get access to Service Management Area

Jelledb:


docs/marketplace.service-management-area.md
```diff
@@ -23,6 +23,14 @@ The Service Management Area can at any time be opened up from the Workspace Cont
 
 If you would like to go back again to the regular Workspace View, click on "Go to Workspace Management".
 
+## Getting Access
+
+If you do not see the button to open the Service Management Area your workspace most likely does not
```
chewbeccy, line 28: The Service Management Area is also its own module. We have a flag in the background das disables the whole area for accounts.

chewbeccy (approved):

PR #1074, opened 2025-05-06
feat: generate llms.txt

felixzieger:


hack/generate-llms-txt
```diff
@@ -0,0 +1,46 @@
...
+echo "Reading markdown files from: $CONTENT_DIR"
+
+# Create output directory if it doesn't exist
+mkdir -p "$OUT_DIR"
+
+# Temporary file to collect all markdown content
+TEMP_FILE=$(mktemp)
+
+# Find all markdown files and concatenate their content
+find "$CONTENT_DIR" -type f -name "*.md" | while read -r file; do
```
sttomm, line 31: d: Would it make sense to also write an llm-toc.txt or similar, that only contains the headlines? Probably it then simply makes sense to only write the llm-toc.txt, which basically provides only the different md files on a top-level with all sub-headlines as "content" and then in the drill down we can get the specific md file. I guess we can just retrieve those files from github directly. That way, we would only generate the toc and don't need that super huge file.

sttomm (changes requested):
I added an idea for a different solution that relies on the drill-down approach

sttomm:
We will probably not go with the llm.txt. At least nit for now, so this PR gets closed

PR #1118, opened 2025-07-01
Update faq.md

AnneHuber1:
With the use of Copilot and Google Gemini we had to update our Subcontractor list.


docs/faq.md
```diff
@@ -149,6 +151,14 @@ Address: 88 Colin P Kelly Jr Street,
          San Francisco,
          California 94107, United States
 ```
+**Google Cloud Gemini**: meshcloud GmbH uses Gemini for processing of user requests to the meshStack Copilot. Use of the service is configured with data storage and processing exclusively in EU regions.
+
+```text
+Gemini Apps Privacy Hub: https://support.google.com/gemini/answer/13594961?hl=en-GB
```
JohannesRudolph, line 157: We use gemini models via google cloud vertex ai (gemini apps may be a different thing). we should list that under GCP

PR #1168, opened 2025-08-01
New guide prompt file

chewbeccy:


.github/instructions/guide.prompt.md
```diff
@@ -0,0 +1,74 @@
...
+   - Use subheadings for major steps if needed.
+   - Include tips, best practices, or warnings in callout blocks (`:::tip`, `:::warning`, etc.).
+
+6. **Related Resources**
+   - Link to relevant concepts, other guides, or documentation for further learning.
+   - Split this in two sections: "Concepts" and "Guides".
+
+## Tone and Style
+
+- Follow the meshStack tone instructions
```
sttomm, line 37: d: Does AI know what the meshStack tone instructions are? Or do you have to describe them additionally? When I read it, I would not know how to achieve this tone

.github/instructions/guide.prompt.md
```diff
@@ -0,0 +1,74 @@
+---
+applyTo: '**/new-guide-*.md'
```
sttomm, line 2: f: You named the file like a prompt, but what you actually have here is an instruction. So you should call the file guide.instructions.md. An instruction is applied also to existing files. A prompt instead is accessible via the slash commands in the copilot chat window. And they are usually used to create new files. So it also makes sense to add a prompt for writing a guide. In the prompt file, which must not be located in the instructions folder, but in the prompts folder, you can then describe what the user needs to provide for actually writing a guide. You could also use tools like executing commands or searching existing files, etc in that prompt file to automatically determine some context. But that might be hard for the guides. So I would keep the prompt a bit more fuzzy, but let the AI make sure that the user provides the most important content for a guide. The header of a prompt then looks like this:

```
---
mode: 'agent'
tools: ['editFiles', 'search']
---
```

You can also find many examples in the meshfed-release repository for it.

docs/new-guide-how-to-manage-a-tenant.md
```diff
@@ -12,7 +12,11 @@ This guide helps you manage tenants in meshStack. You'll learn how to keep your
 ### Prerequisites
 
 - Have an admin user that can log into meshStack.
+<<<<<<< HEAD
```
sttomm, line 15: f: remove the remainder of a merge conflict

.github/instructions/guide.prompt.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Related Resources
+
+- [Relevant Concept](concept-id)
+- [Related Guide](guide-id)
+```
+
+---
+
+Use this prompt as a checklist and template whenever you create or update a guide in
```
sttomm, line 74:
```suggestion
Use this instruction as a checklist and template whenever you create or update a guide.
```

sttomm (changes requested):
There is a little misunderstanding of prompt vs instruction files. I want to have another short look when that is fixed

.github/instructions/guide.prompt.md
chewbeccy: Its a file Johannes already created that is applied to all copilot actions

PR #1172, opened 2025-08-04
docs: updated bb capabilities table with latest features

JohannesRudolph:


JohannesRudolph:
well, as long as it is in the documentation it should be accurate. We can consider culling it in the future when revamping the OSB docs

PR #1241, opened 2025-09-22
Feature/aws and azure hierarchical assignment

sttomm:


docs/integrations/aws/how-to-integrate.md
```diff
@@ -579,7 +592,7 @@ This line item will be added to the report every month until the end of the Rese
 
 To achieve this, meshStack collects the information about Reserved Instances and Savings Plans via the cost explorer API.
 
-### 11. Set Up AWS Tenant Import (Optional)
+### 12. Set Up AWS Tenant Import (Optional)
```
Jelledb, line 595: you might have to check that the links in the panel refer the correct number otherwise the anchor links break

Jelledb (approved):

docs/integrations/aws/how-to-integrate.md
sttomm: uff, that is pretty annoying that we have to update the links when we change the numbering. But I guess the only way to do it atm :/ Here is the PR: https://github.com/meshcloud/meshfed-release/pull/8557

